PR #1, opened 2024-09-08
docs: default Laravel namespace path

balu-lt:
Hey TJ,

Great work on the package!
I'm building AI apps with Laravel, and I can already see how useful this will be.

This is my first pull request, which suggests an update to the README.
I recommend using the default namespace in the examples to reduce unnecessary complexity.

Thanks again for all your effort—really excited to see where this goes!

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,13 +247,13 @@ Agents are registered with the `SparkleServer` facade in a service provider. Her
 ```php
 <?php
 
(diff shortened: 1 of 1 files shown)


sixlive:
Hey! Thanks for the PR! This was a copy and paste artifact form the test suite, good catch!

PR #2, opened 2024-09-08
Update README.md

balu-lt:
I should have double-checked it before the first PR.
Here are a few more that I missed.

I was second-guessing about `datetime`, but the suggested change is a proper phrase for clarity.


sixlive (commented):
Thank you! These docs where definitely a bit hasty, really appreciate it!

PR #6, opened 2024-10-08
Update README.md

veeqtoh:
This change updates the README to show clear examples on how to use the package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,12 +34,32 @@ This will create a new file at `config/prism.php`. We'll explore how to configur
 ```php
 <?php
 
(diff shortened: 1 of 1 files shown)


sixlive (changes requested):
I think adding some more basic examples is fine. 

Could you combine them into 1 code block though?

veeqtoh:
> I think adding some more basic examples is fine.
> 
> Could you combine them into 1 code block though?

Many thanks @sixlive. That's it now combined.

sixlive (approved):

Issue #9, opened 2024-10-09
No OpenAI Organization option


peterramsing:
I'd like to add the ability to add the organization header.

https://platform.openai.com/docs/guides/production-best-practices

Unless this is excluded on purpose I've roughed out a PR and just need to fix the tests (tomorrow, it's too late and I need some sleep right now).

PR #10, opened 2024-10-09
Adds OpenAI Organization to Header

peterramsing:
Adds the ability to pass in an OpenAI Organization to the header

Fixes: #9


src/Drivers/OpenAI/Client.php
```diff
@@ -14,10 +14,12 @@ class Client
 
     public function __construct(
         public readonly string $apiKey,
+        public readonly string $organization,
         public readonly string $url,
     ) {
         $this->client = Http::withHeaders([
```
yoavf, line 20: Why not build the headers array prior to the `withHeaders` call?
Something along the lines of:
```suggestion
$headers = [
    'Authorization' => sprintf('Bearer %s', $this->apiKey),
];

if (!empty($this->organization)) {
    $headers['OpenAI-Organization'] = $this->organization;
}

$this->client = Http::withHeaders($headers)->baseUrl($this->url);
```

peterramsing: Noice! 👏🏼 Thanks @yoavf

src/Drivers/OpenAI/Client.php
```diff
@@ -14,11 +14,19 @@ class Client
...
         public readonly string $url,
     ) {
-        $this->client = Http::withHeaders([
+
+        $headers = [
             'Authorization' => sprintf('Bearer %s', $this->apiKey),
-        ])->baseUrl($this->url);
+        ];
+
+        if ($this->organization !== '' && $this->organization !== '0') {
```
peterramsing, line 25: The formatter keeps adding this. I'm tempted to remove the `!== '0'`.

peterramsing:
Tonight I'll figure out the right way to test this and remove the draft.

tests/Http/PrismHeadersTest.php
peterramsing, line 1: My ignorance is showing here => I wasn't quite sure how to add the check in the existing tooling so I created this `HTTP::fake` => I'm very open to being shown how this _should_ be done. 😄 🚀

src/Drivers/OpenAI/Client.php
```diff
@@ -15,10 +15,12 @@ class Client
     public function __construct(
         public readonly string $apiKey,
         public readonly string $url,
+        public readonly ?string $organization = null,
     ) {
-        $this->client = Http::withHeaders([
-            'Authorization' => sprintf('Bearer %s', $this->apiKey),
-        ])->baseUrl($this->url);
+        $this->client = Http::withHeaders(array_filter([
+            'Authorization' => $this->apiKey !== '' && $this->apiKey !== '0' ? sprintf('Bearer %s', $this->apiKey) : null,
```
peterramsing, line 21: @sixlive I think the formatting got a bit heavy handed here (`composer format did this to mine, too`) but I don't think passing a `API` key is valuable since that will surely `401`. Instead I think it should Raise, but also type-checking won't let a `null` be passed to the API.

src/Drivers/OpenAI/Client.php
```diff
@@ -15,10 +15,12 @@ class Client
...
+            'OpenAI-Organization' => $this->organization,
```
peterramsing, line 22: @sixlive, I think this would then have a blank value to `OpenAI-Organization` if one is not present, right? This should have the conditional re-added.

tests/Drivers/OpenAITextTest.php
```diff
@@ -94,3 +96,51 @@
...
+    FixtureResponse::fakeResponseSequence('v1/chat/completions', 'openai/generate-text-with-a-prompt');
+
+    Prism::text()
+        ->using('openai', 'gpt-4')
+        ->withPrompt('Who are you?')();
+
+    Http::assertSent(fn (Request $request): bool => $request->header('OpenAI-Organization')[0] === 'echolabs');
+});
+
+it('does not sends the organization header', function (): void {
```
peterramsing, line 112:
```suggestion
it('does not send the organization header if one is not given', function (): void {
```

tests/Drivers/OpenAITextTest.php
```diff
@@ -94,3 +96,51 @@
...
+it('sends the organization header when set', function (): void {
+    config()->set('prism.providers.openai.organization', 'echolabs');
+
+    FixtureResponse::fakeResponseSequence('v1/chat/completions', 'openai/generate-text-with-a-prompt');
+
+    Prism::text()
+        ->using('openai', 'gpt-4')
+        ->withPrompt('Who are you?')();
+
+    Http::assertSent(fn (Request $request): bool => $request->header('OpenAI-Organization')[0] === 'echolabs');
```
peterramsing, line 109: @sixlive ahh, these tests are so much better. Thank you.

tests/Drivers/OpenAITextTest.php
```diff
@@ -94,3 +96,51 @@
...
+it('does not sends the organization header', function (): void {
+    config()->offsetUnset('prism.providers.openai.organization');
+
+    FixtureResponse::fakeResponseSequence('v1/chat/completions', 'openai/generate-text-with-a-prompt');
+
+    Prism::text()
+        ->using('openai', 'gpt-4')
+        ->withPrompt('Who are you?')();
+
+    Http::assertSent(fn (Request $request): bool => empty($request->header('OpenAI-Organization')));
```
peterramsing, line 121: I'd have to check, but this sees that the header is empty, but (see https://github.com/echolabsdev/prism/pull/10/files#r1795618106) I think we want to make sure the header isn't present at all.

peterramsing:
@sixlive, the tests are so much cleaner. Thanks for knowing how to wield that `HTTP` better than I!

--
I think the conditional logic for adding it got a bit muddled with both lint added the `''` and `'0'` check, along with always passing the header, though. I have to jump to work, but will check my assumptions on that "always passing the header" tonight or during lunch if I have some extra time.

src/Drivers/OpenAI/Client.php
sixlive: Since we are wrapping it in `array_filter` if there isn't an organization preset and it defaults to a `null` value, the header is removed if there isn't an organization configured.

peterramsing: doh => pre-coffee code reading. Yes. 🤦🏼

src/Drivers/OpenAI/Client.php
peterramsing: Probably out of scope for this PR, but if the API key is optional then we likely should have it be set as optional as well.

tests/Drivers/OpenAITextTest.php
peterramsing: @sixlive , this works as expected. If the header is present, even if it doesn't have a value, then the tests will fail. This is 👍🏼

peterramsing:
@sixlive , this is ready for a final review, unless if we wanted to deal with making the API key optional in the same way that the organization is optional. Not sure if we want to add more to this PR, though.

Issue #3, opened 2024-10-07
Mistral Support


sixlive:
## Description

Add Mistral to the first party supported providers. We should do some investigation but I'm pretty sure all we need to do is add the configuration using the OpenAI driver, similar to Ollama.

https://docs.mistral.ai/api/

pushpak1300:
Can I work on this @sixlive ?

sixlive:
@pushpak1300 yeah! Go for it!

Issue #5, opened 2024-10-07
Investigate a way to expose driver specific features


sixlive:
## Description

We need to figure out a way to expose driver specific features. For example, Anthropic [prompt caching](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching).

## Claude Caching

This is how Vercel implemented Claude caching: https://sdk.vercel.ai/providers/ai-sdk-providers/anthropic#example-cache-control

petervandijck:
(Not sure this is relevant but I thought I'd mention it) Can we use composition to expose underlying features? 
`return $this->api->$method(...$args);`

sixlive:
Closing because this ticket is too open ended, will spin up a specific ticket for prompt caching

PR #13, opened 2024-10-09
feat: add mistral driver

pushpak1300:
This pull request adds the Mistral driver to the codebase. 

Resolves https://github.com/echolabsdev/prism/issues/3


src/Drivers/OpenAI/OpenAI.php
```diff
@@ -68,11 +68,14 @@ public function text(TextRequest $request): DriverResponse
             ));
         }
 
-        $toolCalls = array_map(fn (array $toolCall): ToolCall => new ToolCall(
-            id: data_get($toolCall, 'id'),
-            name: data_get($toolCall, 'function.name'),
-            arguments: data_get($toolCall, 'function.arguments'),
-        ), data_get($data, 'choices.0.message.tool_calls', []));
+        $toolCalls = [];
```
peterramsing, line 71: Did you mean to include this in this PR for Mistral?

src/Drivers/OpenAI/OpenAI.php
peterramsing, line 1: Did you mean to include this in this PR for Mistral?

peterramsing: ...that said, should this have its own driver? ...it might be a bit wet code...should it have a base Driver to extend? 🤔

pushpak1300: No it make sense. Having it's own driver can give flexibility to add mistral related features easily. Will add the driver

sixlive: I think using the OpenAI  driver is fine unless there are some Mistral specific features that we want to expose. 

That being said, I'm still ideating on a way to expose driver/provider specific features outside of the unified API (https://github.com/echolabsdev/prism/issues/5).

src/Drivers/OpenAI/OpenAI.php
```diff
@@ -68,11 +68,14 @@ public function text(TextRequest $request): DriverResponse
...
+        if ($rawToolCalls = data_get($data, 'choices.0.message.tool_calls')) {
+            $toolCalls = array_map(fn (array $toolCall): ToolCall => new ToolCall(
+                id: data_get($toolCall, 'id'),
+                name: data_get($toolCall, 'function.name'),
+                arguments: data_get($toolCall, 'function.arguments'),
+            ), $rawToolCalls);
+        }
```
sixlive, line 78: @pushpak1300 can you explain the reasoning behind this change?

tests/Fixtures/ollama/generate-text-with-multiple-tools-2.json
sixlive, line 1: Can you revert the change to this file?

sixlive (changes requested):
Looking good, have a quick question and a minor change

src/Drivers/OpenAI/OpenAI.php
pushpak1300: @sixlive in case if tool call is not there Mistral API send empty array as a tool property.

tests/Fixtures/ollama/generate-text-with-multiple-tools-2.json
pushpak1300: Done.

src/Drivers/OpenAI/OpenAI.php
sixlive: I think we could do this instead, yeah?

```php
        $toolCalls = array_map(fn (array $toolCall): ToolCall => new ToolCall(
            id: data_get($toolCall, 'id'),
            name: data_get($toolCall, 'function.name'),
            arguments: data_get($toolCall, 'function.arguments'),
        ), data_get($data, 'choices.0.message.tool_calls', []) ?? []);
```

pushpak1300: Pushed the change @sixlive

src/Drivers/OpenAI/OpenAI.php
Darkbelg: The problem is when OpenAi changes their api. Which they have just done for o1 models. max_tokens got deprecated for max_completion_tokens. So If you would update this it would break the mistral driver. But this will be needed if you want to add o1. If each one has it's own driver there is less maintence needed long term i think. Or less thinking about how are we going to handle this upgrade.

sixlive: Thats a really good point! I'm going to simmer on this a bit and see what I come up with. It might make sense to move away from the driver model and just make it 1:1 for provider → driver. I think we would run into this issue a lot with the pace of evolution in the LLM space.

peterramsing: At the very-least, having it named `openai` is confusing if it's used for Ollama, and Mistral, as well. 

Could there be a generic driver that is extended to simplify this? ...and in theory, the generic could be used until features would cause it to split? re: #5

pushpak1300: I guess add this point creating some kind of base driver (let's say) OpenAI Driver and we can extend that driver to use 1 →  1 driver. also there is one extra feature which mistral provides at this point which other provide doesn't eg. OPENAI
https://docs.mistral.ai/capabilities/guardrailing/

sixlive: After sleeping on it an thinking about it some more, I think we probably need to move away from multi-use drivers. I don't think we can expect providers like Mistral, Ollama, and Grok to keep up with the pace of change in OpenAI's API contract. 

I definitely think there is a use case for a OpenAIAPI base driver or a collection of traits to handle the overlap in the driver functionality. 

This weekend I'll work on re-working things for a 1:1 driver → provider with some abstractions. I'll implement the Ollama driver as a use case.

sixlive: I merged the changes for #19 should be good to go to refactor this PR for Mistral support!

pushpak1300: @sixlive Okay Perfect. I have already started working on this.

pushpak1300:
@sixlive this is ready for review.

sixlive:
@pushpak1300 I just enabled CI for the PR (have to approve outside forks). I'll take a look at this tonight!

src/Providers/Mistral/Client.php
```diff
@@ -0,0 +1,49 @@
...
+use Illuminate\Http\Client\Response;
+use Illuminate\Support\Facades\Http;
+
+class Client
+{
+    protected PendingRequest $client;
+
+    public function __construct(
+        public readonly string $url,
+        public readonly ?string $apiKey,
```
sixlive, line 17: We should probably update things so that the API key is required.

src/Providers/Mistral/Client.php
```diff
@@ -0,0 +1,49 @@
...
+    ) {
+        $this->client = Http::withHeaders(array_filter([
+            'Authorization' => $this->apiKey ? sprintf('Bearer %s', $this->apiKey) : null,
+        ]))->baseUrl($this->url);
```
sixlive, line 21: Since the API key is required we can probably remove the conditional here.

src/PrismManager.php
```diff
@@ -81,6 +82,17 @@ protected function createOllamaProvider(array $config): Ollama
         );
     }
 
+    /**
+     * @param  array<string, string>  $config
+     */
+    protected function createMistralProvider(array $config): Mistral
+    {
+        return new Mistral(
+            apiKey: $config['api_key'] ?? '',
```
sixlive, line 91: We should make the API key required, so we can remove the default here.

src/Providers/Mistral/Mistral.php
```diff
@@ -0,0 +1,102 @@
...
+
+class Mistral implements Provider
+{
+    protected Client $client;
+
+    protected string $model;
+
+    public function __construct(
+        public readonly string $url,
+        public readonly ?string $apiKey,
```
sixlive, line 24: We should make the API key required

sixlive (changes requested):
Looks good, just need to make the API key required and we'll be good to go!

src/Providers/Mistral/Mistral.php
pushpak1300: @sixlive  Pushed the changes. [9802f42](https://github.com/echolabsdev/prism/pull/13/commits/9802f423c1b91da29b93098baab58e8e505dd8b5)

src/PrismManager.php
pushpak1300: Pushed the changes. [9802f42](https://github.com/echolabsdev/prism/pull/13/commits/9802f423c1b91da29b93098baab58e8e505dd8b5)

src/Providers/Mistral/Client.php
pushpak1300: Pushed the changes. [9802f42](https://github.com/echolabsdev/prism/pull/13/commits/9802f423c1b91da29b93098baab58e8e505dd8b5)

src/Providers/Mistral/Client.php
pushpak1300: Pushed the changes. [9802f42](https://github.com/echolabsdev/prism/pull/13/commits/9802f423c1b91da29b93098baab58e8e505dd8b5)

tests/Providers/MistralTextTest.php
```diff
@@ -0,0 +1,93 @@
...
+declare(strict_types=1);
+
+namespace Tests\Providers;
+
+use EchoLabs\Prism\Facades\Tool;
+use EchoLabs\Prism\Prism;
+use Tests\Fixtures\FixtureResponse;
+
+beforeEach(function (): void {
+    config()->set('prism.providers.mistral.api_key', env('MISTRAL_API_KEY'));
```
sixlive, line 12:
```suggestion
    config()->set('prism.providers.mistral.api_key', env('MISTRAL_API_KEY', 'sk-1234'));
```

config/prism.php
```diff
@@ -17,5 +17,9 @@
         'ollama' => [
             'url' => env('OLLAMA_URL', 'http://localhost:11434/v1'),
         ],
+        'mistral' => [
+            'api_key' => env('MISTRAL_API_KEY'),
```
sixlive, line 21:
```suggestion
            'api_key' => env('MISTRAL_API_KEY', ''),
```

sixlive (changes requested):
Two quick changes, I left suggestions

sixlive (approved):
Can't tell you how much I appreciate this!! Great work!

PR #22, opened 2024-10-12
Refactor Drivers to Providers

### LINKED ISSUE #19: Refactor drivers and providers
## Description

After further consideration, we need to move away from multi-use drivers. It's becoming apparent that providers like Mistral, Ollama, and Grok may struggle to keep pace with the rapid changes in OpenAI's API contract.

**Proposed Changes:**

- Implement a 1:1 driver → provider model
- Create an OpenAIAPI base driver or a collection of traits to handle overlapping driver functionality
- Rework the existing system to accommodate these changes
- Implement the Ollama driver as a use case for the new model

sixlive:
## Description

Closes #19 

Not completely sold on this implementation, just seeing what it feels like.


sixlive:
https://github.com/echolabsdev/prism/pull/23

Issue #32, opened 2024-10-18
Support for Vision


slashequip:
Current we cannot send user messages that are image URLs or base64 encodes, as outlined in OpenAI's Vision docs;

https://platform.openai.com/docs/guides/vision

I'd expect this could be an easy one to to add, we'd just need a new Message type, and adjusting the message maps appropriately to build the correct payload.

slashequip:
I'll spin up a PR for this and see how far I can get it :)

sixlive:
I'm definitely interested in this, its on the [short-term roadmap](https://prism.echolabs.dev/roadmap.html#short-term-goals)

PR #34, opened 2024-10-19
Draft: Add support for Vision

slashequip:
Adds support for Vision to [OpenAI](https://platform.openai.com/docs/guides/vision) and [Anthropic](https://docs.anthropic.com/en/docs/build-with-claude/vision).


slashequip:
@sixlive what do you think of this approach?

slashequip:
Would resolve #32

sixlive:
I think this is a solid start, I'll sleep on it and might pick this up tomorrow. This is definitely high on my list of things I want to enable.

slashequip:
I've actually rolled almost exactly this logic out in a custom OpenAI provider from my app internally and it works great.

sixlive:
image

I've got it it working with Prism Server for OpenAI but it will need some work for Anthropic. This isn't too far off how I was going to approach things. I'm going to run with this a bit. Its going to be my focus going into next week.

sixlive:
Closing in favor of https://github.com/echolabsdev/prism/pull/42

PR #43, opened 2024-10-24
Handle exception when invalid param passed to a tool

### LINKED ISSUE #39: Exception improvement
I think the below error could be caught and improved to make it a little more clear as to what the issue is.

SCR-20241022-jhht

It results from you defining a Tool with expected parameters and then not including those args in your callable (in my case an `__invoke` method).

Caught me out a few times now when adjusting the Tool parameters and not accounting for them in the handling. The exception makes sense, I think it could be caught and transformed into something more helpful.

pushpak1300:
Resolves #39

### DIFF
--- a/src/Exceptions/PrismException.php
+++ b/src/Exceptions/PrismException.php
@@ -39,6 +39,14 @@ class PrismException extends Exception
         );
     }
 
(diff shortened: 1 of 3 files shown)


sixlive (approved):
Thank you so much!!

PR #49, opened 2024-10-27
Tool choice and required tool selection

### LINKED ISSUE #21: Required tools
We should allow for required tools, meaning the model is required to use a specific tool.

sixlive:
## Description

Closes #21 

```php
$weather = Tool::as('weather')
    ->for('useful when you need to search for current weather conditions')
    ->withStringParameter('city', 'The city that you want the weather for')
    ->using(fn (string $city): string => 'The weather will be 75° and sunny');

Prism::text()
    ->withTools([$weather])
    ->withToolChoice(ToolChoice::Auto)
    ->withToolChoice(ToolChoice::Any)
    ->withToolChoice('weather');
```


mrfelipemartins:
It would be nice to able to pass a tool class like `->withToolChoice(SearchTool::class)`, as described in [Using Tools in Prism](https://prism.echolabs.dev/tool-calling.html#defining-complex-tools)

sixlive:
@mrfelipemartins great suggestion! I added support for passing `Tool` classes as well!

PR #52, opened 2024-11-03
Handle system messages as user for Anthropic

sixlive:
## Description

This PR updates the Anthropic provider to handle `SystemMessages` as a user message rather than throwing.


pushpak1300 (approved):
LGTM !

PR #55, opened 2024-11-09
feat: add xai support

### LINKED ISSUE #50: Grok text completions
Is it possible to use prism with xAI & Grok (not Groq)? It seems to be 'compatible' with OpenAI & Anthropic according to [the xAI documentation](https://docs.x.ai/api/integrations#openai-sdk).

```
import OpenAI from "openai";
const openai = new OpenAI({
  apiKey: "<api key>",
  baseURL: "https://api.x.ai/v1",
});

const completion = await openai.chat.completions.create({
  model: "grok-beta",
  messages: [
    { role: "system", content: "You are Grok, a chatbot inspired by the Hitchhiker's Guide to the Galaxy." },
    {
      role: "user",
      content: "What is the meaning of life, the universe, and everything?",
    },
  ],
});
```
I am confused how to do it though - I can't add another provider from the prism config file and I can't modify the existing provider to use the xAI server & API key.

pushpak1300:
Resolves #50


sixlive:
For context: I've never used Grok/xAI before.

Can you take a look at the providers docs (`docs/providers/xai.md`) and help me fill in the providers page (take a look at [OpenAI](https://github.com/echolabsdev/prism/blob/main/docs/providers/openai.md?plain=1) and the [provider template](https://github.com/echolabsdev/prism/blob/main/docs/providers/providers.md.template) for context).

I want to make sure we capture any specific configurations or limitations. Based on the code, it doesn't look like Grok supports images, is this true of the API?

pushpak1300:
Consideration wise XAI only support Text Generation and Tool calling. Image support is not there ate the moment.I will update the docs accordingly.

pushpak1300:
@sixlive I have added this is docs. Please check it.

sixlive (approved):

PR #65, opened 2024-11-18
feat: anthropic cache support

pushpak1300:


sixlive:
I was thinking about handling it through something like we handle the provider specific options at the top level and through tools but add it to messages and prompts.

image

However, I'm open to ideas.

ChrisB-TL:
@sixlive I haven't dug into how withProviderMeta is used yet, but my assumption is it is currently for "request level" stuff?

If so, it would need to work slightly differently for Anthropic prompt caching as it requires a header + a flag at the message level. And for most use cases you'd be applying it to one message but not others. E.g. if putting a book in as context, you'd cache the book text, but not the user's questions about the book.

ChrisB-TL:
> @sixlive I haven't dug into how withProviderMeta is used yet, but my assumption is it is currently for "request level" stuff?
> 
> If so, it would need to work slightly differently for Anthropic prompt caching as it requires a header + a flag at the message level. And for most use cases you'd be applying it to one message but not others. E.g. if putting a book in as context, you'd cache the book text, but not the user's questions about the book.

I've looked at how Tool uses withProviderMeta, and can see how this would work for messages. Am I right in thinking the API for cached messages would therefore be the following?

```
Prism::structured()
    ->using(Provider::Anthropic, 'claude-3-5-sonnet-20241022')
    ->withMessages(
        (new UserMessage('My cached prompted'))->withProviderMeta(Provider::Anthropic, ['cache_control' => ['type' => 'ephemeral']])
        (new UserMessage('My uncached prompt'))
    )
```

Would also need to think about the API for a system prompt, which also supports caching.

PR #75, opened 2024-11-30
Add nullable schema types

### LINKED ISSUE #69: Nullable Schema types
Hi, thanks for the awesome library!

I was testing structured outputs and found a limitation.

[Because all fields must be required in the schema](https://platform.openai.com/docs/guides/structured-outputs#all-fields-must-be-required), OpenAI recommends making the type a nullable union if one wants an "optional" field.

```diff
{
    "name": "get_weather",
    "description": "Fetches the weather in the given location",
    "strict": true,
    "parameters": {
        "type": "object",
        "properties": {
            "location": {
                "type": "string",
                "description": "The location to get the weather for"
            },
            "unit": {
+               "type": ["string", "null"],
                "description": "The unit to return the temperature in",
                "enum": ["F", "C"]
            }
        },
        "additionalProperties": false,
        "required": [
            "location", "unit"
        ]
    }
}
```

As far as I understand, this is not possible with the current Schema objects, right? Would be nice if it could be added in the future. :)

cosmastech:
To close #69 (nice)

### DIFF
--- a/src/Schema/EnumSchema.php
+++ b/src/Schema/EnumSchema.php
@@ -16,8 +16,7 @@ class EnumSchema implements Schema
         public readonly string $description,
         public readonly array $options,
         public readonly bool $nullable = false,
-    ) {
-    }
+    ) {}
 
     #[\Override]
     public function name(): string
--- a/src/Schema/EnumSchema.php
+++ b/src/Schema/EnumSchema.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace EchoLabs\Prism\Schema;
 
 use EchoLabs\Prism\Contracts\Schema;
-use Illuminate\Support\Collection;
 
 class EnumSchema implements Schema
 {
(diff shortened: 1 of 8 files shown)


src/Schema/EnumSchema.php
```diff
@@ -29,6 +30,37 @@ public function toArray(): array
...
+            ->unique()
+            ->values()
+            ->all();
+
+        if (count($types) === 1) {
+            return $types[0];
+        }
+
+        return $types;
+    }
```
cosmastech, line 65: This feels a little bit ugly. Happy to take any suggestions you may have!

src/Schema/ObjectSchema.php
```diff
@@ -18,6 +21,7 @@ public function __construct(
         public readonly array $properties,
         public readonly array $requiredFields = [],
```
cosmastech, line 22: Not sure if we should spin through the properties and see which are nullable here? That may break backwards compatibility? Not entirely sure.

sixlive:
> To close https://github.com/echolabsdev/prism/issues/69 (nice) 

nice

src/Schema/EnumSchema.php
sixlive: Maybe it would just be cleaner if we make it a single collection pipeline?

sixlive (commented):
Looks like its headed in the right direction, added a few comments/thoughts

src/Schema/EnumSchema.php
```diff
@@ -29,6 +31,31 @@ public function toArray(): array
...
+            'type' => $this->getOptionTypes(),
         ];
     }
+
+    /**
+     * @return string|array<int, string>
+     */
+    protected function getOptionTypes(): array|string
+    {
+        // @phpstan-ignore return.type
```
cosmastech, line 43: Without this, I'm getting this error:

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Schema/EnumSchema.php                                                                                                                                                                
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  43     Method EchoLabs\Prism\Schema\EnumSchema::getOptionTypes() should return array<int, string>|string but returns array<int, string>|Illuminate\Support\Collection<int, string>|string.  
         🪪  return.type                                                                                                                                                                      
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

I would personally probably forego the last Collection::when() call, since I don't think it makes it easier to read, and is throwing static analysis for a loop.

```php
$types = collect($this->options)
    ->map(fn (mixed $option): string => match (gettype($option)) {
        'integer', 'double' => 'number',
        'string' => 'string'
    })
    ->when(
        $this->nullable,
        fn (Collection $collection): Collection => $collection->push('null')
    )
    ->unique()
    ->values();

if ($types->count() === 1) {
    return $types[0];
}

return $types->all();
```

Let me know your thoughts @sixlive.

cosmastech:
Ready for another look @sixlive

sixlive:
@cosmastech What do you think of the refactors I just pushed to the EnumSchema?

cosmastech:
> @cosmastech What do you think of the refactors I just pushed to the EnumSchema?

Works for me. 👍

sixlive:
Awesome, just stepped out but I'll merge a little later today! 

Really appreciate it 💜

sixlive (approved):

PR #72, opened 2024-11-27
Mistral Embeddings

pushpak1300:
This MR Adds Support For Mistral Embeddings.

Resolves #76

PR #85, opened 2024-12-04
Refactor EmbeddingsResponse to Support Multiple Embeddings

ThaKladd:
Updates the embeddings endpoint usage to handle responses containing multiple embeddings from the OpenAI API, as this is how it works as of [OpenAi Embeddings docs](https://platform.openai.com/docs/api-reference/embeddings). Previously, the implementation only supported a single embedding, and it adds network overhead when dealing with multiple chunks.

What changed:

- fromInput method on Prism::embeddings() accepts array as input
- $response->embeddings is always an array of arrays, meaning if a text is sent in, to retrieve it would be in $response->embeddings[0]

Considerations:

- I considered the possibility of adding a first() method, so that it works similar as to Laravels eloquent, but decided to follow the standard set forth by existing code


pushpak1300:
I'm not sure if @sixlive wants to support that yet.
https://github.com/echolabsdev/prism/pull/72#pullrequestreview-2470374651

ThaKladd:
Ah. Now I saw the comment made on the Mistral Embeddig PR. 😅 Should have looked closer.

sixlive:
I don't know, if this is the second time bringing it up, maybe I should reconsider 🤔

sixlive:
I'm going to close this for now but I'll be willing to revisit it in the future. Thanks!!

PR #101, opened 2024-12-12
Anthropic: allow for multiple system prompts

ChrisB-TL:
Most providers' API include system messages in the messages array with a "system" role, and allow you to make multiple. Anthropic does not support the system role, and instead has a "system" property, separate from messages.

Prism currently allows for one system prompt via `->withSystemPrompt()`, and converts all `SystemMessage`s into `UserMessage`s. 

It is not well documented, but the system property allows for an array of system prompts. See for instance [the example in the prompt caching docs](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching). I have validated that this works.

This PR:
* Filters out all `SystemMessage`s from the main map, therefore omitting them from messages.
* Moves the system property to the array of prompts format.
* Always submits the prompt defined with `->withSystemPrompt()` at the top of the system prompts array (same behaviour as other providers implementations, which puts it at the top of the messages array).
* Move all `SystemMessage`s to the system prompts array in the order they were declared.

**Why is this needed?**

Primarily because Anthropic prompt caching supports system prompts but requires you to use the array format to specify the cache_control property. This is therefore a precursor to implementing prompt caching.

**Why not just move to an array format, with a single system prompt?**

* If your system prompt has a cacheable part, and un uncacheable part, you need to split it into two prompts.
* When it comes to implementing prompt caching, by treating `SystemMessage`s correctly as system prompts, you would be able to use the same API for both system prompts and user prompts. E.g. (making some assumptions about the desired API):

```
->withMessages([
    (new SystemMessage('Cache me'))->withProviderMeta(Provider::Anthropic, ['cache_control' => ['type' => 'ephemeral']]),
    (new UserMessage('Cache me too'))->withProviderMeta(Provider::Anthropic, ['cache_control' => ['type' => 'ephemeral']]),
    new UserMessage('Please do not cache me')
])
```

**Is this a breaking change?**

Kind of, but not really! 

If you were using SystemMessages before, they would have been submitted as UserMessages in the order that they were defined (i.e. potentially between UserMessages). SystemMessages would now still be sent in the correct order compared to each other, but are taken out of the context of the UserMessages.

However, given the docs currently state that all SystemMessages are converted to UserMessages, I cannot imagine many (if any) people are doing this.

This is also really a limitation of Anthropic's API implementation, which doesn't allow you to place system prompts between other prompts.

### DIFF
--- a/docs/providers/anthropic.md
+++ b/docs/providers/anthropic.md
@@ -27,7 +27,12 @@ While Anthropic models don't have native JSON mode or structured output like som
 ## Limitations
 ### Messages
 
-Does not support the `SystemMessage` message type, we automatically convert `SystemMessage` to `UserMessage`.
+Most providers' API include system messages in the messages array with a "system" role. Anthropic does not support the system role, and instead has a "system" property, separate from messages.
+
+Therefore, for Anthropic we:
+* Filter all `SystemMessage`s out, omitting them from messages.
+* Always submit the prompt defined with `->withSystemPrompt()` at the top of the system prompts array.
+* Move all `SystemMessage`s to the system prompts array in the order they were declared.
 
 ### Images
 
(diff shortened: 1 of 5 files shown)


ChrisB-TL:
Have had to close this to clean up our fork. Will resubmit when done.

PR #104, opened 2024-12-13
Anthropic prompt caching

ChrisB-TL:
Implements Anthropic prompt caching

**Choices made**
* Ended up opting for ->withProviderMeta() on UserMessage, SystemMessage and AssistantMessage - (1) to keep consistent with Tools, and (2) as it felt a bit clumsy as a parameter on the constructor due to the "provider" wrapper (e.g. `providerMeta: [Provider::Anthropic->value -> [...]]`. It ended up easy enough to do with a trait, shared with Tool.
* Have used a "beta_features" config rather than prompt caching, to future proof and allow for enabling other beta features (e.g. PDF processing).

I don't think providerMeta is the nicest API for the user, and it might get messy in the map methods if more features become reliant on it. But it should be easy to add a nicer API in a backward compatible way in future if needed. My gut thought is a ->withFeature() method on Messages and Tools, which accepts an implementation of a contract with a handle method, which are then applied with a pipeline to mutate the payload.

**To-do**
* Once the multiple system prompts PR is merged, fix mapSystemMessage() and write a test for it.
* Update docs.

P.S. could do away with the `AnthropicCacheType` enum - however if anyone else is like me that can never spell "ephemeral" right - it is quite handy!


src/Providers/Anthropic/Enums/AnthropicCacheType.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace EchoLabs\Prism\Providers\Anthropic\Enums;
+
+enum AnthropicCacheType
+{
+    case ethemeral;
```
pushpak1300, line 7: I see value of keeping the enum here but Imo keeping provider related settings in the prisms code can make maintenance way harder. 

@sixlive can think different.

src/Providers/Anthropic/Maps/MessageMap.php
```diff
@@ -86,10 +91,16 @@ protected static function mapUserMessage(UserMessage $message): array
...
+        $cacheType = data_get($message->providerMeta(Provider::Anthropic), 'cacheType', null);
+
         return [
             'role' => 'user',
             'content' => [
-                ['type' => 'text', 'text' => $message->text()],
+                array_filter([
+                    'type' => 'text',
+                    'text' => $message->text(),
+                    'cache_control' => $cacheType ? ['type' => $cacheType instanceof UnitEnum ? $cacheType->name : $cacheType] : null,
```
pushpak1300, line 102: Instead of setting cache_control as null. can we not append the `cache_control` key only when caching is enabled.

ChrisB-TL: Thanks for reviewing! Array filter removes the null value, following the pattern in e.g. [OpenAI ToolMap](https://github.com/echolabsdev/prism/blob/main/src/Providers/OpenAI/Maps/ToolMap.php).

pushpak1300: yep make sense. I missed it.

src/Providers/Anthropic/Enums/AnthropicCacheType.php
ChrisB-TL: Hilariously, having added the enum to avoid spelling mistakes, I misspelt it in the enum...! Now fixed.

Definitely not a huge loss if we don't include the enum. Really just a convenience thing.

ChrisB-TL:
Prompt caching is now out of beta.

The only live beta is now computer use, so we can probably just bin the config/header changes - unless you'd rather leave it in for future @sixlive?

ChrisB-TL:
Have had to close this to clean up our fork. Will resubmit when done.

PR #116, opened 2025-01-09
[WIP] Gemini Structured Outputs

btumbleson:
Hey folks, this is a bit beyond my comfort zone, but wanted to share my WIP code for structured outputs with Gemini. 

This code is working for me with a handful of somewhat complex schemas (I've explicitly used/ tested the `Object`, `Array`, `String` and `Number` schemas, with nesting), although I acknowledge I haven't attached any tests here (yet).

Some observations that may need to be addressed more meaningfully:

- The `ObjectSchema` defines an `allowAdditionalProperties` key which is valid for OpenAI, but not for Gemini. As the schema itself isn't aware of the `Provider` being chosen, without doing some injection/ facade/ whatever, it was easier to just strip this key out after the `toArray()` map. There may need to be a longer term decision made here based on how other providers treat their schema definitions, or perhaps a provider-specific `map()` or `toArray()` handler for the `schema`.
- Gemini doesn't appear to have multiple structured options like OpenAI, so the `StructuredModeResolver` is a bit superfluous. Granted it's still a valid check on if a specific model supports structured mode, but I skipped this logic inclusion in my WIP pass here as all "current" Gemini models (excluding v1 Pro) support structured outputs.

Thoughts on a path forward to formalizing this?

### DIFF
--- a/src/Providers/Gemini/Gemini.php
+++ b/src/Providers/Gemini/Gemini.php
@@ -9,6 +9,7 @@ use EchoLabs\Prism\Embeddings\Request as EmbeddingRequest;
 use EchoLabs\Prism\Embeddings\Response as EmbeddingResponse;
 use EchoLabs\Prism\Providers\Gemini\Handlers\Embeddings;
 use EchoLabs\Prism\Providers\Gemini\Handlers\Text;
+use EchoLabs\Prism\Providers\Gemini\Handlers\Structured;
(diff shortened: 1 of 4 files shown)


sixlive:
Hey! I'm sorry that I never got to this, just got buried in other things. Going to close this in favor of #173. 

I really appreciate the effort!!

PR #140, opened 2025-01-20
fix: correct method name for tool choice in documentation

ianfortier:
Fix incorrect method name from ->toolChoice() to ->withToolChoice() in the documentation


sixlive (approved):
Thank you so much!!

PR #142, opened 2025-01-20
Anthropic PDF support

ChrisB-TL:
A relatively simple PR to add support for [Anthropic PDF processing](https://docs.anthropic.com/en/docs/build-with-claude/pdf-support).

The only thing of note from an architecture POV is the addition of a `Document` value object and corresponding method on `UserMessage`. As they are currently only used for Anthropic, I've left a note in there as such!


sixlive (approved):
This is SICK! Thanks!

Issue #141, opened 2025-01-20
More granular exceptions


ChrisB-TL:
I am happy to PR this, but wanted to check before I do whether it would be better to wait - e.g. if the areas in question are about to be refactored.

At the moment, we throw a `PrismException` for all exceptions, which is fine for development.

However, I'd like to add specific exceptions extending `PrismException` for more common errors - e.g. `PrismRateLimitedException` for rate limits - so that we can easily handle them by e.g. backing off. 

Any objections or notes before I put together a PR? 

In particular, as I wouldn't test against all providers live, do you think we are safe to assume all return proper error codes - i.e. a 429 for rate limits?

sixlive:
No pending refactors in this area, so you're clear to proceed! 

Re: provider consistency - you raise a good point about status codes. While we can't guarantee all providers follow HTTP conventions perfectly (429s etc.), I think introducing specific exceptions is a solid iterative improvement. Start with the providers you can verify and we can enhance coverage over time.

Go ahead with the PR! Focus on the providers where you have high confidence in the error handling behavior. We can always circle back later to expand coverage, or let other community members contribute provider-specific enhancements as they encounter edge cases.

pushpak1300:
like the idea. 

Are you working on other exception such as 401 auth errors or 404 error `PrismInvalidApiKeyException` ?

ChrisB-TL:
Yes, though I was going to prioritise those that are likely to come up unexpectedly in production for monitoring and alerting. E.g. insufficient credit, overloaded, request too large, etc. which tend to be a bit trickier as providers don't handle them all in the same way.

That said, once we have an approach nailed - 401 and 404, 500 etc should be very quick/easy to add. Given those are probably universal across all providers and exclusively error code driven, we can probably just come up with a single shared "inspector".

PR #144, opened 2025-01-23
feat: rate limit exceptions and Anthropic implmentation 

ChrisB-TL:
Re: #141 

### What does this PR do?

1. Refactors Anthropic handlers to extend an abstract base handler - to reduce duplication going forward - particularly as we'll possibly have more handlers imminently (e.g. token counting).
2. Creates a `RateLimitException`, which accepts an array of `ProviderRateLimit`s for the user to inspect.
3. Implements throwing the `RateLimitException` for Anthropic.
4. Adds an array of `ProviderRateLimit`s to `ResponseMeta` for Anthropic.
5. Adds an optional headers param to `FixtureResponse::fakeResponseSequence()` to support testing (4).

I have done (1) as a separate commit, in case you'd rather split into two PRs.

### Notes

- I have added an empty `PrismRequest` interface for strong typing on the abstract class. 
- I have moved the handlers' request and http response to properties for ease of accessing and type hinting. 

### Questions

How would you like to cover this in the docs? A new Rate Limits page under Core Concepts, with explanation of the exception, ResponsaMeta, and maybe a provider support table?

### Breaking changes

I don't think there should be any, other than where people have extended the Anthropic handler classes.

### DIFF
--- a/src/Providers/Anthropic/Handlers/Structured.php
+++ b/src/Providers/Anthropic/Handlers/Structured.php
@@ -4,42 +4,57 @@ declare(strict_types=1);
 
 namespace EchoLabs\Prism\Providers\Anthropic\Handlers;
 
-use EchoLabs\Prism\Exceptions\PrismException;
+use EchoLabs\Prism\Contracts\PrismRequest;
 use EchoLabs\Prism\Providers\Anthropic\Maps\FinishReasonMap;
 use EchoLabs\Prism\Providers\Anthropic\Maps\MessageMap;
-use EchoLabs\Prism\Structured\Request;
+use EchoLabs\Prism\Structured\Request as StructuredRequest;
 use EchoLabs\Prism\ValueObjects\Messages\UserMessage;
 use EchoLabs\Prism\ValueObjects\ProviderResponse;
 use EchoLabs\Prism\ValueObjects\ResponseMeta;
 use EchoLabs\Prism\ValueObjects\Usage;
 use Illuminate\Http\Client\PendingRequest;
-use Illuminate\Http\Client\Response;
-use Throwable;
 
-class Structured
+class Structured extends AnthropicHandlerAbstract
 {
+    /**
+     * @var StructuredRequest
+     */
+    protected PrismRequest $request; // Redeclared for type hinting
+
     public function __construct(protected PendingRequest $client) {}
 
-    public function handle(Request $request): ProviderResponse
+    /**
+     * @param  StructuredRequest  $request
+     * @return array<string, mixed>
+     */
+    #[\Override]
+    public static function buildHttpRequestPayload(PrismRequest $request): array
     {
-        try {
-            $request = $this->appendMessageForJsonMode($request);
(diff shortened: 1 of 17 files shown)


ChrisB-TL:
Latest commit fixes the structured message not being added (didn't realise addMessage returns a new request).

We need to add a test for this, which will be do-able once #146 is merged. I'll rebase to main once it has and add the test.

ChrisB-TL:
@sixlive I didn't think that through - we still can't test for it as the json mode message is added after toRequest - which steps builds from.

I think we leave this test out for now, until we add the raw request body to ProviderResponse, StructuredResponse, etc. - on my todo list once I have sorted out our rate limit issues!

ChrisB-TL:
I've now added docs to this. Revisiting, it does need some decent explanation so I think it warrants its own docs page?

src/Providers/Anthropic/Handlers/Structured.php
```diff
@@ -4,42 +4,60 @@
...
-        try {
-            $request = $this->appendMessageForJsonMode($request);
-            $response = $this->sendRequest($request);
-            $data = $response->json();
-
-        } catch (Throwable $e) {
-            throw PrismException::providerRequestError($request->model, $e);
+        if (! $request instanceof StructuredRequest) {
+            throw new \InvalidArgumentException('Request must be an instance of '.StructuredRequest::class);
         }
```
sixlive, line 35: Could we maybe abstract this to a trait? `$request->is(Structured::class)` or something like that?

src/Providers/Anthropic/Handlers/Text.php
```diff
@@ -4,43 +4,56 @@
...
     {
-        try {
-            $response = $this->sendRequest($request);
-            $data = $response->json();
-
-        } catch (Throwable $e) {
-            throw PrismException::providerRequestError($request->model, $e);
+        if (! $request instanceof TextRequest) {
+            throw new \InvalidArgumentException('Request must be an instance of '.TextRequest::class);
         }
```
sixlive, line 32: Could we maybe abstract this to a trait? `$request->is(Structured::class)` or something like that?

sixlive:
> I've now added docs to this. Revisiting, it does need some decent explanation so I think it warrants its own docs page?

I think these docs are solid! I can't even express how much I appreciate the extensive docs, means a lot to me.

pushpak1300 (approved):

src/Providers/Anthropic/Handlers/Structured.php
ChrisB-TL: I think I have done what you mean - but let me know if not!

P.S. I couldn't help but call the trait "ChecksSelf"! Once it was in my head, I couldn't think of an alt...

sixlive (approved):
Love this, the Anthropic handler abstract is great. I think this is going to help with some refactors down the line!

sixlive:
Closing in favor of #174. Things got weird fixing the merge conflicts.

PR #145, opened 2025-01-23
feat: Add getters to ConfiguresProviders

ChrisB-TL:
A micro PR to add getters to `ConfiguresProviders` - for convenience where they have been dynamically set and you need to apply some logic based on the `PendingRequest`.

### DIFF
--- a/src/Concerns/ConfiguresProviders.php
+++ b/src/Concerns/ConfiguresProviders.php
@@ -41,4 +41,19 @@ trait ConfiguresProviders
 
         return $this;
     }
+
+    public function getModel(): string
+    {
+        return $this->model;
+    }
+
+    public function getProviderKey(): string
+    {
+        return $this->providerKey;
+    }
+
+    public function getProviderEnum(): ProviderEnum
+    {
+        return ProviderEnum::from($this->providerKey);
+    }
 }
--- a/src/Concerns/ConfiguresProviders.php
+++ b/src/Concerns/ConfiguresProviders.php
@@ -42,18 +42,13 @@ trait ConfiguresProviders
         return $this;
     }
 
-    public function getModel(): string
+    public function model(): string
     {
         return $this->model;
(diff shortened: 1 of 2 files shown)


src/Concerns/ConfiguresProviders.php
```diff
@@ -41,4 +41,19 @@ public function usingProviderConfig(array $config): self
...
+
+    public function getProviderKey(): string
+    {
+        return $this->providerKey;
+    }
+
+    public function getProviderEnum(): ProviderEnum
+    {
+        return ProviderEnum::from($this->providerKey);
+    }
```
sixlive, line 58: I'm thinking that we only return the key, since both can be used. There will ALWAYS be a provider key, but not always a provider Enum because users can register custom providers that aren't going to be in the Enum.

```suggestion
    public function providerKey(): string
    {
        return $this->providerKey;
    }
```

tests/Text/PendingRequestTest.php
```diff
@@ -39,6 +39,15 @@
         ->toHaveKey('openai', ['key' => 'value']);
 });
 
+test('it allows you to get the model and provider', function (): void {
+    $request = $this->pendingRequest
+        ->using(Provider::OpenAI, 'gpt-4');
+
+    expect($request->getModel())->toBe('gpt-4');
+    expect($request->getProviderKey())->toBe('openai');
+    expect($request->getProviderEnum())->toBe(Provider::OpenAI);
```
sixlive, line 48: Based on the other suggestions...

```suggestion
    expect($request->model())->toBe('gpt-4');
    expect($request->providerKey())->toBe('openai');
```

sixlive (commented):
Left a few comments/suggestions

ChrisB-TL:
@sixlive all make sense - thank you!

sixlive (approved):

PR #146, opened 2025-01-23
fix: PendingRequest refactor regressions

ChrisB-TL:
### Regression 1 - exception thrown if you use `toRequest()` more than once

If you run `toRequest()` on a `PendingRequest` `before generate()` - e.g. to inspect the request before running it - you trigger a `promptOrMessage` exception as it populates the messages property on the first run.

Fixed by manipulating a variable rather than the property.

### Regression 2 - Text Generator not adding system prompt to first step

Previously, a system prompt would be included in the first step's messages array. This PR reinstates it.

### Regression 3 - Structured Generator not adding system prompt or prompt to first step

Similar to above, but also applies to the user prompt.


ChrisB-TL:
I've noticed a couple of other regressions from the PendingRequest refactor, so I have marked this as draft and will add fixes to this same PR.

sixlive (approved):
Thank you so much!! That refactor has bit me a few times now but I think it was worth it!

PR #150, opened 2025-01-24
feat: OpenAI Project Header

Perturbatio:
* added support for OpenAI Project Header (OpenAI-Project)
* (see https://platform.openai.com/docs/api-reference/authentication)


sixlive (approved):
Thanks!!!

PR #153, opened 2025-01-26
feat(anthropic): citations

ChrisB-TL:
Note: I have drafted this on the assumption #144 gets merged. If it doesn't, or it does with changes, I'll need to re-work this a little.

### What does this PR do?

This required a fair bit of fiddling, but I managed to do it without changing APIs so much. I've pinched the concept of additionalContent from messages - I think this is the best way to do it acknowledging that other providers may never support citations - or if they do they could look quite different.!

- Adds `array $additionalContent = []` to `ProviderResponse`, `StructuredResponse`, `TextResponse`, `Step` and `AssistantMessage`.
- Adds `messagePartsWithCitations` to `$additionalContent`, which is an array of `Providers\Anthropic\ValueObjects\MessagePartWithCitations`;
- Allows you to enable by either using `->withProviderMeta(Provider::Anthropic, [‘citations’ => true])` on either:
  - each document; or
  - or the `PendingRequest`, which will apply to all `Document`s on `UserMessage`s.

At the moment, Anthropic only supports it on all or no documents, so the latter is a shortcut to adding it on every document. I’ve allowed the Document API as I suspect this may be supported by Anthropic in future, and it avoid a refactor when the time comes!


ChrisB-TL:
This should now be done, but I've left marked as WIP until #144.

I have managed to get this to working with structured by adding to the JSON mode prompt, though it is a little prickly. I've left a warning as such in the docs.

I've only documented enabling citations at the request level for now - as the JSON mode prompt adjustment depends on it.

ChrisB-TL:
Have rebased to main. #163 caused me some pain - I have hopefully sorted all of the issues.

Should now be ready for review.

xHeaven:
> Have rebased to main. #163 caused me some pain - I have hopefully sorted all of the issues.
> 
> Should now be ready for review.

The `Document` class can be readonly, other than that it looks fine to me.

ChrisB-TL:
Had to remove readonly from document as HasProviderMeta brings in a none readonly property.

xHeaven:
> Had to remove readonly from document as HasProviderMeta brings in a none readonly property.

Ah, you're right. I've missed that one.

sixlive (approved):
This is sick!! Thanks @ChrisB-TL !!

PR #154, opened 2025-01-27
fix: reinstate assistant message to steps

ChrisB-TL:
This one is on me - in #146 I re-instated the system and user message to the first step, but accidently knocked off the assistant message. Oh the irony!

Have fixed + expanded the test for text and structured to cover this.


sixlive (approved):
Should have caught this! Thanks for the quick fix!

Issue #158, opened 2025-01-30
System message duplicated using structured outputs


mdcass:
I'm attempting and failing to get structured outputs from `llama3.2:latest`, and finding an interesting different issue.

* The `\EchoLabs\Prism\Providers\Ollama\Maps\MessageMap` constructor is duplicating the system message of the following code example
* No response is returned from the local model and a `PrismException` is returned

```php
$responseSchema = new ObjectSchema(
    name: 'response_schema',
    description: 'A response to a journal entry',
    properties: [
        new StringSchema('response', 'The response to the journal entry'),
        new ArraySchema(
            name: 'follow_up_questions',
            description: 'Follow up questions to the journal entry which might provoke insightful reflection',
            items: new ObjectSchema(
                name: 'question',
                description: 'A follow up question to the journal entry which might provoke insightful reflection',
                properties: [
                    new StringSchema('question_content', 'The question'),
                ],
                requiredFields: ['question_content'],
            ),
        ),
    ],
    requiredFields: ['response', 'follow_up_questions'],
);

$response = Prism::structured()
    ->using(Provider::Ollama, 'llama3.2:latest')
    ->withSchema($responseSchema)
    ->withSystemPrompt('You are a micro journal response agent. Input will be a journal entry. You should respond to input with praise and constructive comments. However you think you are better than everyone, respond in a condescending tone, respond with eccentricity, your responses are aloof, and slightly indifferent - to the point of irreverent humour. Answers should be succinct and no longer than two paragraphs.')
    ->withPrompt('I woke up today feeling refreshed and I did yoga.')
    ->generate(); // PrismException
```

sixlive:
Thanks for reporting! I'm in the middle of re-writing the Ollama provider (#157). I'll make sure to address this as part of the refactor.

mdcass:
Thanks for the prompt reply!

Some more debugging, within `\EchoLabs\Prism\Structured\PendingRequest::toRequest` a `new SystemMessage` is added to the `$messages` array, and is passed through to a `new Request` along with a `systemPrompt`. Then when we eventually get to `\EchoLabs\Prism\Providers\Ollama\Handlers\Structured::sendRequest:64` which compiles a message using `\EchoLabs\Prism\Providers\Ollama\Maps\MessageMap`, the constructor here is appending the system prompt again to the messages where it's already been done before.

ChrisB-TL:
[...]

PR #161, opened 2025-02-01
Fix duplication of systemPrompt using withSystemPrompt

ChrisB-TL:
This is a fix for issue #158, which was caused by PR #154.

Per my message on the issue:

- We do need to add the system message to messages in `Request` in order for it to be available on the response, so we cannot remove it without a refactor of generator.
- $prompt is not used at all, other than in testing assertions (easy fix)
- $systemPrompt is only used by handlers to add it to the top of the messages array anyway.

This PR removes $prompt and $systemPrompt from `Request`s, and refactors $systemPrompt out of handlers. This leaves $prompt and $systemPrompt as a "pure generator API thing", as a shortcut for `withMessages`.

This fixes the issue, simplifies handlers, and hopefully mitigates the impact of any future refactors to the core (i.e. non-provider) handling. 

Given that - `toRequest()` now throws an exception if trying to use `withSystemPrompt()` and `withMessages()` together, to be consistent with `withPrompt`.

### Breaking changes
- Can no longer use `withSystemPrompt` and `withMessages` together.
- If folks were using $prompt or $systemPrompt - e.g. in a customer provider. We could leave them on `Request` and just not use them to avoid this, but long term I suspect its better to just remove these to avoid edge cases. I don't think handlers should ever need them outside of the messages array (and if they do, they could just pluck them out).

All tests are passing, and I cannot think of any edge cases. However @sixlive you have a much wider understanding, so welcome any thoughts!

### DIFF
--- a/docs/core-concepts/text-generation.md
+++ b/docs/core-concepts/text-generation.md
@@ -49,18 +49,23 @@ $response = Prism::text()
 You an also pass a View to the `withPrompt` method.
 
 ## Message Chains and Conversations
+> [!NOTE]
+> You should avoid using `withSystemPrompt()` for message chains. Doing so will result in that SystemMessage being duplicated in all but the first generation.
+> Instead, include your system prompt as a `SystemMessage` at the top of your first `withMessages()` call.
 
 For interactive conversations, use message chains to maintain context:
 
 ```php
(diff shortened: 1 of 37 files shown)


ChrisB-TL:
Rebased to main and fixed merge conflicts.

PR #162, opened 2025-02-01
feat: New Provider - Cohere

nea:
Ahoy there

First of all thanks for this awesome package.

I would like to propose the addition of [Cohere](https://cohere.com/) as provider, which I think provides good service and excellent models.

The given implementation suggests basic embeddings, text and structured handling, similar to Mistral as both share similar APIs.
Some specialties such as `response_format` for structured output have been added.
The implementation follows the APIv2 spec, as it is standard for about 6 months now and the differences to v1 are bigger.
Tests have been added in all conscience.

I hope this PR fits the package and standards. Let me know if you look for changes.

Cheers


nea:
Hey there

If there is something I should change or add to have this added to Prims, please let me know :) 


Thanks and Best

sixlive:
I'm going to close this as its going to require a fair bit of work to update after a lot of internal changes. I'd be happy to accept a new PR for cohere!

PR #165, opened 2025-02-03
fix(openai): function definitions without parameters

mattmcdonald-uk:
OpenAI supports functions that don't take any parameters.

If you pass an empty array of parameters an error is returned:
```Sending to model (model-name) failed: OpenAI Error:  [invalid_request_error] Invalid schema for function 'tool_name': [] is not of type 'object'.```

You instead don't send `parameters` when defining a function without arguments.


sixlive (approved):
Nice!! Great catch! Thank you so much!

PR #170, opened 2025-02-06
Support null value maxTokens

mattmcdonald-uk:
This PR defaults `maxTokens` to 2048 to keep the same behaviour as today, whilst respecting developer requests to set value to `null`.


ChrisB-TL:
+1! Also makes the fact that the package defaults to 2048 slightly more discoverable (kinda!) and consistent across providers.

pushpak1300 (approved):
Looks good to me. This should get in.

sixlive (approved):

PR #176, opened 2025-02-10
fix(gemini): support tool calling without params

mattmcdev:
In the same vein as https://github.com/echolabsdev/prism/pull/165 this PR allows calling tools with Gemini that don't require parameters

### DIFF
--- a/src/Providers/Gemini/Maps/MessageMap.php
+++ b/src/Providers/Gemini/Maps/MessageMap.php
@@ -124,7 +124,9 @@ class MessageMap
             $parts[] = [
                 'functionCall' => [
                     'name' => $toolCall->name,
-                    'args' => $toolCall->arguments(),
+                    ...count($toolCall->arguments()) ? [
+                        'args' => $toolCall->arguments(),
+                    ] : [],
                 ],
             ];
         }
--- a/src/Providers/Gemini/Maps/ToolMap.php
+++ b/src/Providers/Gemini/Maps/ToolMap.php
@@ -21,11 +21,13 @@ class ToolMap
         return array_map(fn (Tool $tool): array => [
             'name' => $tool->name(),
             'description' => $tool->description(),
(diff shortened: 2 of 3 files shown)


src/Providers/Gemini/Maps/MessageMap.php
```diff
@@ -124,7 +124,9 @@ protected function mapAssistantMessage(AssistantMessage $message): void
             $parts[] = [
                 'functionCall' => [
                     'name' => $toolCall->name,
-                    'args' => $toolCall->arguments(),
+                    ...count($toolCall->arguments()) ? [
+                        'args' => $toolCall->arguments(),
+                    ] : [],
```
pushpak1300, line 129: I don't think this is needed. 
LLM directly sends empty array in case no parameter is needed.

mattmcdonald-uk:
```
Prism::text()
    ->using(Provider::Gemini, 'gemini-2.0-flash')
    ->withTools(
        [
            FacadesTool::as('weather')
                ->for('Get the current weather for the user.')
                ->using(fn (): string => 'It is sunny.'),
        ]
    )
    ->withMessages([
        new UserMessage('What is the current weather for the user ?'),
    ])
    ->withMaxSteps(2)
    ->generate()
```

If you run this without that change you'll get the error: `Gemini Error: [400] Invalid JSON payload received. Unknown name "args" at 'contents[1].parts[0].function_call': Proto field is not repeating, cannot start list.`

pushpak1300: Okay let me try this. I tried with gemini 1.5 flash.

pushpak1300 (approved):

pushpak1300:
@mattmcdonald-uk i tested this approach without Tool Choice and it threw the error. 

I can confirm it works with and without tool choice now.

PR #173, opened 2025-02-08
Feature/gemini structured output support

dev-billy:
Hey everyone! 👋
This is my first PR to Prism, so I’m pretty excited (and a little nervous 😅). I’ve added structured output support for the Gemini provider using a prompt-based approach, similar to how it’s done for Anthropic and Deepseek Providers.

### What’s in this PR?
- Adds structured output for Gemini using prompt formatting.
- Includes tests to make sure everything works as expected.

### Why this approach?

I noticed another PR [#116](https://github.com/echolabsdev/prism/pull/116) exploring structured output for Gemini with a schema-based approach. My take is a little different—more in line with how it is being handled for Anthropic. The only difference with anthropic, gemini api config allows for `response_mime_type` which when set to `application/json` ensures the response is a JSON 

### Would love your thoughts!

Since this is my first contribution, I’d really appreciate any feedback! Let me know if this makes sense or if there’s anything I can tweak.

Thanks for taking a look! Looking forward to your thoughts. 🚀😊

PR #177, opened 2025-02-10
feat(gemini): support structured outputs

mattmcdev:
Created this PR before realising [this one](https://github.com/echolabsdev/prism/pull/173) existed.

The main difference is this PR updates the structured output to match the schema format Google defines. It then passes this directly using `response_schema` rather than appending a message.

To me this seems the preferred way of doing it:
- The schema matching the defined format is going to help the model know how to use your schema.
- Passing the schema via `response_schema` means you get validation (and relevant errors) on your schema

### DIFF
--- a/docs/components/ProviderSupport.vue
+++ b/docs/components/ProviderSupport.vue
@@ -198,7 +198,7 @@ export default {
         {
           name: "Gemini",
           text: Supported,
-          structured: Planned,
+          structured: Supported,
           embeddings: Supported,
           image: Supported,
           tools: Supported,
--- a/docs/providers/gemini.md
+++ b/docs/providers/gemini.md
@@ -7,7 +7,3 @@
     'url' => env('GEMINI_URL', 'https://generativelanguage.googleapis.com/v1beta/models'),
 ],
 ```
-
-## Limitations
-
-- The structured output is not supported.
--- a/src/Providers/Gemini/Gemini.php
+++ b/src/Providers/Gemini/Gemini.php
@@ -9,6 +9,7 @@ use EchoLabs\Prism\Embeddings\Request as EmbeddingRequest;
 use EchoLabs\Prism\Embeddings\Response as EmbeddingResponse;
 use EchoLabs\Prism\Exceptions\PrismException;
 use EchoLabs\Prism\Providers\Gemini\Handlers\Embeddings;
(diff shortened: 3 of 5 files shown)


ChrisB-TL:
I'm not yet familiar with the Gemini API but if it accepts a proper schema that would seem the better route.

Do you know if there limitations at all, which would make the Anthropic adaption approach better?

We use the Anthropic adaption a lot, and whilst it works most of the time it isn't completely reliable. It goes off piste around 5% of the time.

mattmcdonald-uk:
> I'm not yet familiar with the Gemini API but if it accepts a proper schema that would seem the better route.
> 
> Do you know if there limitations at all, which would make the Anthropic adaption approach better?
> 
> We use the Anthropic adaption a lot, and whilst it works most of the time it isn't completely reliable. It goes off piste around 5% of the time.

[Google's docs](https://ai.google.dev/gemini-api/docs/structured-output) seem to encourage using the schema via model configuration:
 > Important: This more formal method for declaring the JSON schema gives you more precise control than relying just on text in the prompt.

The only potential (sort of) downside is your call will fail if the schema doesn't pass validation which wouldn't happen with the Anthropic approach - personally I'd prefer to know that my schema was invalid and correct it.

ChrisB-TL:
Yea, I see that it would fail at the request level as an advantage - you don't get billed for the tokens for a broken request/response.

pushpak1300:
I think we should support both the implementation as both are supported by gemini  at the moment. https://github.com/echolabsdev/prism/pull/173 similar like openai structred support.

`src/Providers/OpenAI/Handlers/Structured.php`

ziming:
Very interested in this too! This is especially more so now that Gemini is leading in OCR

pushpak1300 (approved):

ianfortier:
I've just tested it! Works great!

ianfortier (approved):

sixlive:
I'm ready to merge this but we need to updated it based on https://github.com/echolabsdev/prism/pull/177

sixlive (changes requested):
Needs to be updated based on https://github.com/echolabsdev/prism/issues/185 but other than that this looks good to go!

sixlive (approved):
Thank you so much!

PR #178, opened 2025-02-11
feat(openai): support required tool calling

mattmcdev:
Prism docs mention a limitation that OpenAI doesn't support `ToolChoice::Any`, but there's a mappable option of `required` as per [their docs](https://platform.openai.com/docs/guides/function-calling#additional-configurations)

Using this option as-is has the same problem as using a named tool - it will loop forever using it.

This PR proposes respecting that tool setting for the first call, and reverting to the OpenAI default of `auto` for subsequent calls.

### DIFF
--- a/docs/providers/openai.md
+++ b/docs/providers/openai.md
@@ -34,8 +34,3 @@ $response = Prism::structured()
         ] // [!code focus]
     ]) // [!code focus]
 ```
-
-## Limitations
-### Tool Choice
-
-OpenAI does not support `ToolChoice::Any` when using `withToolChoice()`.
--- a/src/Concerns/ConfiguresTools.php
+++ b/src/Concerns/ConfiguresTools.php
@@ -10,12 +10,15 @@ use EchoLabs\Prism\Tool;
 trait ConfiguresTools
 {
     protected string|ToolChoice|null $toolChoice = null;
+    protected int $toolChoiceAutoAfter = 1;
 
-    public function withToolChoice(string|ToolChoice|Tool $toolChoice): self
+    public function withToolChoice(string|ToolChoice|Tool $toolChoice, int $toolChoiceAutoAfter = 1): self
     {
         $this->toolChoice = $toolChoice instanceof Tool
             ? $toolChoice->name()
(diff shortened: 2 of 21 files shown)


pushpak1300:
@mattmcdonald-uk I think this is the intedned beahvior the only change we need here is

`src/Providers/OpenAI/Maps/ToolChoiceMap.php`

```php
return match ($toolChoice) {
            ToolChoice::Auto => 'auto',
            ToolChoice::Any => 'required', //this
            null => $toolChoice,
            default => throw new InvalidArgumentException('Invalid tool choice')
        };
 ```
 
 and removing the limitation from docs. That is suffiecient. 
 
whenever you use ToolChoice::Any you should use with max steps to avoid getting loop.


@sixlive let me know if this make sense.

mattmcdonald-uk:
> @mattmcdonald-uk I think this is the intedned beahvior the only change we need here is
> 
> `src/Providers/OpenAI/Maps/ToolChoiceMap.php`
> 
> ```
> return match ($toolChoice) {
>             ToolChoice::Auto => 'auto',
>             ToolChoice::Any => 'required', //this
>             null => $toolChoice,
>             default => throw new InvalidArgumentException('Invalid tool choice')
>         };
> ```
> 
> and removing the limitation from docs. That is suffiecient.
> 
> whenever you use ToolChoice::Any you should use with max steps to avoid getting loop.
> 
> @sixlive let me know if this cis correct or I am misisng anything.

The difference is that using `ToolChoice::Auto` the model may or may not use the tool, but will eventually return a response to your prompt assuming you've set a high enough `maxSteps`. Setting to `required` as is today, the model will loop infinitely until `maxSteps` is hit.

Maybe the behaviour shouldn't be implicit as it would be in this PR ? Maybe it should be that you can define how many steps you want the tool usage to be required for ?

I think there's definitely a big development time-saving advantage in being able to have it automatically run further step(s) after being required to use the tool, to give similar behaviour you get today with `ToolChoice::Auto` where with a high enough `maxSteps` you'll get a response to your prompt, just with the benefit of being certain a tool will be called to inform that response.

PR #179, opened 2025-02-12
fix(tools): catch unknown named parameter errors

mattmcdonald-uk:
While debugging failed tool calls, I was finding it hard to understand quickly why `Calling tool failed` was being thrown.

It was due to the model inventing named parameters to send as an argument to the function.

This change helps expose this problem quicker by:
- Catching PHPs `Unknown named parameter` errors as an additional `invalidParameterInTool` error.
- Throwing a previous `PrismException` rather than `toolCallFailed` when the tool fails.


sixlive (approved):
Thanks!

PR #180, opened 2025-02-12
feat: overloaded and too large exceptions

ChrisB-TL:
This PR:

- Adds exceptions for overloaded and request too large.
- Implements them for Anthopic.
- Adds an exceptions/error handling section to the docs with a provider support table.

I did have a look around at other providers. Not all of them implement all of these exceptions - and if they do they all implement them differently.

I have therefore left a call for contributions in the docs! Hopefully as more people move to production we'll see these coming in.


sixlive (approved):
Sick! Thanks dude!

PR #182, opened 2025-02-12
fix(ollama): embeddings response

uchm4n:
PR fixes Ollama embeddings response: 

- Instead of `usage.total_tokens` response has `prompt_eval_count`
- No need for `data.0.embedding`
- `EmbeddingsUsage` ValueObject has union types `int` and `null`
- Fixed tests fixtures accordingly


sixlive (approved):
thank you so much!!

PR #183, opened 2025-02-12
Fix Ollama embeddings

gdsrmygdsrjr:
The right way: [Ollama Generate Embeddings](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-embeddings)


gdsrmygdsrjr:
closed due to #182

PR #184, opened 2025-02-13
feat: add provider meta to embeddings request

ChrisB-TL:
Absence come across whilst implementing Bedrock Embeddings.


src/Concerns/AccessesProviderMeta.php
sixlive, line 1: Can we just use the `HasProviderMeta` trait?

ChrisB-TL: I did think that - but unfortunately syntax is different on `TextRequest` and `StructuredRequest`, and the `HasProviderMeta` has a protected property vs the existing requests' public.

Though, we are pre 1.0 release, so maybe fine to change that!

ChrisB-TL: (Let me know if you'd rather refactor to use HasProviderMeta and we can mark as breaking.)

sixlive: Maybe we should refactor embeddings to be like the rest? Text and Structured should be the same though.

ChrisB-TL: We'd have to refactor Text and Structured to use HasProviderMeta (with different syntax and property visibility) to match Embeddings (which currently doesn't have provider meta).

sixlive: Oh wait! I see what the issue is here! Sorry, I overlooked that this was on the Request and not on the PendingRequest

ChrisB-TL: Yea, though now thinking about it I think this makes the most sense to keep things uniform going forward, without (much) of a breaking change. Only breaking change then would be that TextRequest and StructuredRequest's ProviderMeta prop goes from public to protected.

> Actually - I could change HasProviderMeta to have an optional second param to match what is on Text and Structured that if not provided falls back to previous behaviour?

ChrisB-TL: image

sixlive: Why do we need the extra check if `$valuePath` is `null`? That should just return the whole array.

ChrisB-TL: Right you are - hadn't appreciated data_get did that! Change incoming.

sixlive: Just re-request review when its ready!

ChrisB-TL:
What do you think of this?  A little messy to make work, but I think the outcome is better...

Had to take readonly back off the classes, but I was not a fan of that anyway!

Alternative is we flip it, and make the trait prop public.

sixlive:
Totally think this is the ticket! Thanks!!!

sixlive (approved):

Issue #185, opened 2025-02-14
Provider/Generator Refactor


sixlive:
# Provider Response Type Refactor

## To Do

- [x] Update Providers
- [x] Remove Generators / Update Provider Contract
- [x] `Prism::fake` updates
- [x] Update docs
  - [x] Custom Provider Docs
  - [x] Testing docs with `Prism::fake`

## Overview

This refactor aims to improve Prism's architecture by moving request handling and response generation logic from the Generator classes to the individual providers. This change will make Prism more flexible and allow providers to implement their own request/response flow in a way that best suits their specific needs.

## Current Implementation

Currently, our Provider interface uses a generic `ProviderResponse` type:

```php
interface Provider
{
    public function text(TextRequest $request): ProviderResponse;
    public function structured(StructuredRequest $request): ProviderResponse;
    public function embeddings(EmbeddingsRequest $request): EmbeddingsResponse;
}
```

The Generator classes are responsible for:
- Managing the request loop
- Invoking tools
- Appending tool results
- Iterating additional calls
- Building the final response

This creates a rigid structure where all providers must conform to the same request/response pattern, even when their APIs might work differently.

## Proposed Changes

We'll update the Provider interface to use specific response types:

```php
interface Provider
{
    public function text(TextRequest $request): TextResponse;
    public function structured(StructuredRequest $request): StructuredResponse;
    public function embeddings(EmbeddingsRequest $request): EmbeddingsResponse;
}
```

### Benefits

1. Providers can implement their own request/response flow
2. More flexible architecture for different API patterns
3. Clearer separation of concerns

### Implementation Strategy

To make this transition manageable, we'll:

1. Update the Provider interface to temporarily accept union types:
```php
public function text(TextRequest $request): TextResponse|ProviderResponse;
```

2. Create new response types for each operation type:
- `TextResponse`
- `StructuredResponse`

3. Migrate providers one at a time to return the new response types

4. Update Generator classes to handle both response types during transition

5. Once all providers are migrated, remove union types and old response types

## Help Needed

This is a major refactor that will touch many parts of the codebase. We need help with:

1. Updating individual providers
2. Writing tests for new response types
3. Updating documentation
4. Code review and testing

### Getting Started

Pick a provider to migrate:

1. Create a fork and new branch `refactor/provider-responses`
2. Update the provider to return the new response type
3. Add tests
4. Submit a PR

### Testing Requirements

- All existing tests should pass
- New tests should cover the new response types
[...]

Issue #160, opened 2025-01-31
System message placement with message chains


KeithBush:
When trying to use the `->withSystemPrompt()` function alongside the `->withMessages()` function, I noticed some unexpected behavior.  (I haven't used this library in any other way yet, so this may not be limited to message chain interactions)

First, when I get a response back and inspect `$response->messages`, it looks like the system prompt got added to the messages array after the message(s) I set via `->withMessages()`.  (Expected: If present, the SystemMessage should always be the first message)

Second, when continuing a conversation the system message is added to the end of the message chain causing it to appear multiple times.  (Expected: if present, the SystemMessage should always be the first message and should not appear more than once)

Note: Provider being used is OpenAI
```php
// $messages is an array of objects implementing the Message interface
// It starts as either an empty array or the stored output from a previous interaction (ie, `$messages = $response->messages->toArray()`)
// Then the latest users prompt gets added to the end (ie, `$messages[] = new UserMessage($prompt)`)

$response = Prism::text()
  ->withSystemPrompt($systemPrompt)
  ->withMessages($messages)
  ->generate();

// When inspecting the order of `$response->messages`, the passed `$messages` appear to be first, followed by the `$systemPrompt`, 
// and then finally the assistant response. 

// Example showing order of initial prompt
$messages = $response->messages->toArray();
[
  0 => EchoLabs\Prism\ValueObjects\Messages\UserMessage, // initial user prompt
  1 => EchoLabs\Prism\ValueObjects\Messages\SystemMessage, // **out of order system message**
  2 => EchoLabs\Prism\ValueObjects\Messages\AssistantMessage, // response to initial prompt
]

// Taking that stored message array and passing it to a second pass when the user submits a new prompt
$messages = $response->messages->toArray();
[
  0 => EchoLabs\Prism\ValueObjects\Messages\UserMessage, // initial user prompt
  1 => EchoLabs\Prism\ValueObjects\Messages\SystemMessage, // **out of order system message**
  2 => EchoLabs\Prism\ValueObjects\Messages\AssistantMessage, // response to initial prompt
  3 => EchoLabs\Prism\ValueObjects\Messages\UserMessage, // second user prompt
  4 => EchoLabs\Prism\ValueObjects\Messages\SystemMessage, // **repeated system message**
  5 => EchoLabs\Prism\ValueObjects\Messages\AssistantMessage, // response to second prompt
]

// What I expected to see after two interactions:
[
[...]

PR #189, opened 2025-02-16
refactor: replace systemPrompt with systemPrompts

### LINKED ISSUE #158: System message duplicated using structured outputs

ChrisB-TL:
In preparation for #185, as discussed I have replaced systemPrompt with systemPrompts.

I have done this on top of #188. If that changes, I can rebase this to main once that is merged.

It was a bit fiddley to do, but I think the outcome is worth it. Much less confusing and much more flexible for the providers which will be handy come provider refactor.

It comes with two relatively minor breaking changes:

- Anthropic no longer supports `SystemMessage`s in the `messages` array. The exception instructs users to use `withSystemPrompt` or `withSystemPrompts` instead.
- `SystemMessage`s will no longer appear in `Step->messages`. They will appear in `Step->systemPrompts` instead.

All other providers behavior remains in tact. Where a provider (e.g. Ollama) only accepts one "separate" system instruction, we throw an exception to let the dev know if more than one is provided.

Fixes #158. Relevant to to #160.

### DIFF
--- a/src/Providers/Groq/Handlers/Text.php
+++ b/src/Providers/Groq/Handlers/Text.php
@@ -63,7 +63,7 @@ class Text
             'chat/completions',
             array_merge([
                 'model' => $request->model(),
-                'messages' => (new MessageMap($request->messages(), $request->systemPrompt() ?? ''))(),
+                'messages' => (new MessageMap($request->messages(), $request->systemPrompts()))(),
                 'max_tokens' => $request->maxTokens ?? 2048,
             ], array_filter([
                 'temperature' => $request->temperature(),
--- a/src/Providers/XAI/Handlers/Text.php
+++ b/src/Providers/XAI/Handlers/Text.php
@@ -63,7 +63,7 @@ class Text
             'chat/completions',
             array_merge([
                 'model' => $request->model(),
-                'messages' => (new MessageMap($request->messages(), $request->systemPrompt() ?? ''))(),
+                'messages' => (new MessageMap($request->messages(), $request->systemPrompts()))(),
                 'max_tokens' => $request->maxTokens ?? 2048,
             ], array_filter([
                 'temperature' => $request->temperature(),
--- a/src/Providers/XAI/Handlers/Text.php
+++ b/src/Providers/XAI/Handlers/Text.php
@@ -64,7 +64,7 @@ class Text
             array_merge([
                 'model' => $request->model(),
                 'messages' => (new MessageMap($request->messages(), $request->systemPrompts()))(),
-                'max_tokens' => $request->maxTokens ?? 2048,
+                'max_tokens' => $request->maxTokens() ?? 2048,
             ], array_filter([
                 'temperature' => $request->temperature(),
                 'top_p' => $request->topP(),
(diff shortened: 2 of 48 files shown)


src/Providers/XAI/Handlers/Text.php
```diff
@@ -62,14 +62,14 @@ public function sendRequest(Request $request): Response
         return $this->client->post(
             'chat/completions',
             array_merge([
-                'model' => $request->model,
-                'messages' => (new MessageMap($request->messages, $request->systemPrompt ?? ''))(),
+                'model' => $request->model(),
+                'messages' => (new MessageMap($request->messages(), $request->systemPrompts()))(),
                 'max_tokens' => $request->maxTokens ?? 2048,
```
uchm4n, line 67: I think you missed the parentheses here. it should be `$request->maxTokens()`

src/Providers/Groq/Handlers/Text.php
```diff
@@ -62,14 +62,14 @@ public function sendRequest(Request $request): Response
         return $this->client->post(
             'chat/completions',
             array_merge([
-                'model' => $request->model,
-                'messages' => (new MessageMap($request->messages, $request->systemPrompt ?? ''))(),
+                'model' => $request->model(),
+                'messages' => (new MessageMap($request->messages(), $request->systemPrompts()))(),
                 'max_tokens' => $request->maxTokens ?? 2048,
```
uchm4n, line 67: Here too.

ChrisB-TL:
@sixlive have rebased to main

pushpak1300:
@sixlive  This MR is gonna cause lot of conflicts in another MR's if you are planning to merge it we should get this in before everything.

sixlive:
@pushpak1300 yeah, I'm digging into this now, definitely going to cause some pain but I think its worth it

sixlive (approved):

PR #191, opened 2025-02-16
refactor(mistral): Return FInal Response

pushpak1300:
Based on https://github.com/echolabsdev/prism/issues/185 I have refactored  mistral provider.
@sixlive can you please review and let me know if this is the correct one and I'll proceed with the all the other one which I have mentioned.


pushpak1300:
I just saw this https://github.com/echolabsdev/prism/pull/188/files looks like i have to wait this to get merged to use new methods instead of props

PR #193, opened 2025-02-16
refactor(ollama): Return Final Response

sixlive:
## Description

Refactor Ollama for #185


src/Providers/Ollama/Handlers/Text.php
```diff
@@ -4,70 +4,155 @@
...
-            throw PrismException::providerResponseError(sprintf(
-                'Ollama Error: %s',
-                data_get($data, 'error', 'unknown'),
-            ));
+    /**
+     * @param  array<string, mixed>  $data
+     */
+    protected function handleToolCalls(array $data, Request $request): Response
+    {
+        if (! $this->shouldContinue($request)) {
```
ChrisB-TL, line 90: I think this changes behavior slightly. 

In the Generator implementation, the step is added before this check, so the step count at this point was previously 1 but is now 0. The generator therefore stops on the first tool call.

Not sure which is the correct behavior - should maxSteps === 1 mean no call tools (which is what Generator would currently do), or one tool call (as here)?

I think one tool call makes sense (and therefore this is right and the generator implementation is wrong)?

Discovered as I borrowed this pattern for Anthropic, and it caused a test failure.

sixlive: Good call!

sixlive: See https://github.com/echolabsdev/prism/pull/193/commits/add6b4183674016e8de552386a2b343c17620942

src/Concerns/DecodesObjects.php
```diff
@@ -0,0 +1,20 @@
...
+declare(strict_types=1);
+
+namespace EchoLabs\Prism\Concerns;
+
+trait DecodesObjects
+{
+    /**
+     * @return array<mixed>|null
+     */
+    public function decodeObject(string $responseText): ?array
```
ChrisB-TL, line 12: This does match current behavior and isn't this implementation specific, but I was wondering on this the other day whether there is a cleaner way to handle this - given that the response builder decodes it again.

E.g. should we just pass this as empty, and allow the response builder to add it to the step after its decoded it?

Or allow only validate here (therefore throwing the new `PrismStructuredDecidingException`, and "trust" it in response builder?

src/Providers/Ollama/Handlers/Text.php
```diff
@@ -4,94 +4,145 @@
...
+    protected function sendRequest(Request $request): array
     {
         try {
-            $response = $this->sendRequest($request);
-            $data = $response->json();
+            $response = $this
+                ->client
+                ->post('api/chat', [
+                    'model' => $request->model(),
+                    'system' => $request->systemPrompt(),
```
ChrisB-TL, line 73: If merging multiple system prompts before, change to:

```
data_get($request->systemPrompts(), '0.content', '')
```

src/Providers/Ollama/Handlers/Text.php
```diff
@@ -4,94 +4,145 @@
...
+            FinishReason::Stop => $this->handleStop($data, $request),
+            default => throw new PrismException('Ollama: unknown finish reason'),
+        };
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    protected function sendRequest(Request $request): array
     {
```
ChrisB-TL, line 67: If merging system prompts before, consider throwing an exception if multiple are provided:

```
        if (count($request->systemPrompts()) > 1) {
            throw new PrismException('Ollama does not support multiple system prompts using withSystemPrompt / withSystemPrompts. However, you can provide additional system prompts by including SystemMessages in with withMessages.');
        }
```

src/Providers/Ollama/Handlers/Structured.php
```diff
@@ -4,64 +4,98 @@
...
+            additionalContent: [],
+        ));
     }
 
-    public function sendRequest(Request $request): Response
+    /**
+     * @return array<string, mixed>
+     */
+    protected function sendRequest(Request $request): array
     {
```
ChrisB-TL, line 82: if (count($request->systemPrompts()) > 1) {
            throw new PrismException('Ollama does not support multiple system prompts using withSystemPrompt / withSystemPrompts. However, you can provide additional system prompts by including SystemMessages in with withMessages.');
        }

src/Providers/Ollama/Handlers/Structured.php
```diff
@@ -4,64 +4,98 @@
...
-            'stream' => false,
-            'options' => array_filter([
-                'temperature' => $request->temperature(),
-                'num_predict' => $request->maxTokens() ?? 2048,
-                'top_p' => $request->topP(),
-            ])]);
+        try {
+            $response = $this->client->post('api/chat', [
+                'model' => $request->model(),
+                'system' => $request->systemPrompt(),
```
ChrisB-TL, line 86: If merging multiple system prompts before, change to:

```
data_get($request->systemPrompts(), '0.content', '')
```

src/Providers/Ollama/Handlers/Structured.php
```diff
@@ -4,64 +4,98 @@
...
                 data_get($data, 'eval_count', 0),
             ),
-            finishReason: FinishReasonMap::map(data_get($data, 'done_reason', '')),
             responseMeta: new ResponseMeta(
                 id: '',
                 model: $request->model(),
-            )
-        );
+            ),
+            messages: $request->messages(),
```
ChrisB-TL, line 73: if merging system prompts before, add:

```
systemPrompts: $this->systemPrompts,
```

src/Providers/Ollama/Handlers/Text.php
```diff
@@ -4,94 +4,145 @@
...
+            toolResults: $toolResults,
+            usage: new Usage(
+                data_get($data, 'prompt_eval_count', 0),
+                data_get($data, 'eval_count', 0),
+            ),
+            responseMeta: new ResponseMeta(
+                id: '',
+                model: $request->model(),
+            ),
+            messages: $request->messages(),
```
ChrisB-TL, line 144: If merging system prompts before, add:

``` 
systemPrompts: $this->systemPrompts,
```

ChrisB-TL (commented):
This looks solid. All but one of these comments are re: additions needed if merging system prompts before.

The other isn't specific to this provider re: whether we can avoid decoding objects once in handler and again in responseBuilder.

src/Concerns/DecodesObjects.php
sixlive: We probably should just let the Response builder handle it, it can decode on the fly when a step is added.

I think I'd rather not throw and just leave the empty array so you can inspect the `text` field if necessary.

sixlive: I removed `object` from `Structured/Step` and just let the response builder decode it to the final response

PR #195, opened 2025-02-17
Set minimum php version to 8.2

da-mask:
Unfortunately peckphp/peck has a minimum version of php8.3, so could not be included. I could find no other reason not have the minimum version set to 8.2.

### DIFF
--- a/docs/input-modalities/images.md
+++ b/docs/input-modalities/images.md
@@ -26,6 +26,17 @@ $message = new UserMessage(
     [Image::fromUrl('https://example.com/diagram.png')]
 );
 
+// From a URL which does not end in the image format,
+// you can pass the mime type as the second argument
+// e.g. if you are generating a temporary URL
+$message = new UserMessage(
+    'Analyze this diagram:',
+    [Image::fromUrl(
+        'https://storage.example.com/diagram.png?AccessID=test&Expires=1742330260&Signature=dVQaFcIk9FJWIVnvV1%2FWu',
+        'image/png'
+    )]
+);
+
 // From a Base64
 $image = base64_encode(file_get_contents('/path/to/image.jpg'));
 


pushpak1300:
can you please add https://github.com/echolabsdev/prism/blob/main/.github/workflows/tests.yml#L18 php8.2 in test matrix so we can be sure if the code works on php 8.2

da-mask:
@pushpak1300 Yes, of course, that makes sense. Do you think the other formatting/phpstan workflows should also be set to 8.2 ?

pushpak1300:
@da-mask I don't think that is needed. Let's wait for @sixlive and see what he thinks ?

sixlive:
I'm okay with this change, I think we can adopt Laravel minimums. I'll take a closer look this afternoon

sixlive:
I think we also need to update the formatting GH Action to remove the peck check as well.

da-mask:
I do have a [PR in with Peck](https://github.com/peckphp/peck/pull/134) to do similar, maybe we give it a few days to see if there is movement over there?

sixlive:
Yeah, lets hold off a few days and see what happens

da-mask:
I just asked Nuno on his twitch stream about the likelihood of approving a php8.2 PR to Peck, and he didn't seem into it. I've added a commit to remove it from the workflow...

da-mask:
Hmm, maybe keep holding on this. A maintainer over at Peck has given the thumbs up on my PR there, and suggested it get merged in. That hasn't happened yet, so I would have to re-add Peck back in here. Let me know if you'd prefer me to create a new PR to keep things clean, or happy for me to add another commit bringing back Peck if/when that PR gets merged.

da-mask:
I closed this as I am waiting for[ a PR](https://github.com/peckphp/peck/pull/139) to be merged and a new version of Peck to be tagged that will have min requirement of php8.2 . I will create a new PR once released.

PR #196, opened 2025-02-18
docs: correct namespace for ProviderResponse

petersuhm:
The correct namespace for ProviderResponse is `EchoLabs\Prism\ValueObjects\ProviderResponse`.


sixlive (approved):
Thanks! I moved it at some point and forgot to update the docs

PR #202, opened 2025-02-22
refactor(openai): Return Final Response

sixlive:
## Description

Refactor OpenAI for #185


src/Providers/Ollama/Handlers/Structured.php
```diff
@@ -21,7 +20,6 @@
 
 class Structured
 {
-    use CallsTools;
```
sixlive, line 24: Shouldn't have been there in the first place.

PR #204, opened 2025-02-23
refactor(gemini): return final Response

ChrisB-TL:
Refactor Gemini for #185.

Note, I have allowed FinishReason::Length as a valid stop reason. I am not sure what the behaviour is in practice with Ollama and OpenAI (i.e. does it still provide a valid response) - but we may wish to consider adding to those also.


sixlive (approved):
Thanks!!!

PR #206, opened 2025-02-23
refactor(groq): return final Response

ChrisB-TL:
Refactor Groq for #185.

Per the others, have allowed FinishReason::Length as a valid stop reason.


sixlive (approved):
Than you!!!

PR #214, opened 2025-02-25
feat: Streaming Ouput & OpenAI Streaming Response

sixlive:
## Description

Streaming output starting with OpenAI.


benbjurstrom:
Pulled this into my local to play around a bit. Working great with OpenAI!

Had a question about how you would go about persisting the output to a database. I've been messing around with a prism powered chatbot and wrote a service to translate the Prism\Text\Response to an eloquent model and back again. It's been working really well for maintaining context in long message chains.

Do you think it will be possible to yield a regular Prism\Text\Response object at the end of the stream?

sixlive:
@benbjurstrom I'll consider it, streaming is already kind of complex so I'm a little hesitant to add more complexity to it. Its something I'll probably circle back to after the initial implementation.

docs/core-concepts/streaming-output.md
```diff
@@ -0,0 +1,120 @@
...
+}
+
+echo "Final response: " . $fullResponse;
+```
+
+## Configuration Options
+
+Streaming supports the same configuration options as regular [text generation](/core-concepts/text-generation#generation-parameters).
+
+## Handling Streaming in Web Applications
```
ChrisB-TL, line 91: Is it worth adding a note along the lines of the below?

```
Alternatively, you might consider using Laravel's [Broadcasting feature](https://laravel.com/docs/11.x/broadcasting) to send the chunks to your frontend.
```

src/Providers/OpenAI/Handlers/Stream.php
```diff
@@ -0,0 +1,246 @@
...
+        $line = trim(substr($line, strlen('data: ')));
+
+        if (Str::contains($line, 'DONE')) {
+            return null;
+        }
+
+        try {
+            return json_decode($line, true, flags: JSON_THROW_ON_ERROR);
+        } catch (Throwable $e) {
+            throw PrismException::providerResponseError(
```
ChrisB-TL, line 109: This feels like an error I'd want to catch, and handle gracefully.

Worth adding a specific exception for this?

ChrisB-TL (commented):
This looks great - a lot of work! Have two really minor suggestions.

PR #219, opened 2025-03-01
feat(gemini): context caching

ChrisB-TL:
## Description

Implements context caching for Gemini - handy for multiple calls with the same long document, image, etc.

Gemini requires you to submit a separate request to cache content, and then you refer to that cached content by name in your request.

It's therefore the first feature in Prism that will require a a method outside the common provider methods (I.e. text, structured, embeddings, stream).

We'll have to cross a similar hurdle as and when we implement e.g. OpenAI document handling (requires separate upload) or Anthropic token counting.

I'd love your thoughts @sixlive - particularly around the API:
- I've assumed we want to be able to do this via the `Prism::` entrypoint
- The way I've done it works, though I am not stoked about the IDE support (I suspect there is a clever way around this...).
- Is this how we want to do it going forward for other provider specific non-message endpoints?

We can use this one as a test case as to how it should be done going forward.


ChrisB-TL:
Have now rebased this to main, so should be cleaner to review.

ChrisB-TL:
Note: will conflict with #239. I will rebase and rectify whichever gets merged second.

sixlive (approved):
I think this seems good!

sixlive:
This should be good to go if you can resolve the conflicts for me!

ChrisB-TL:
@sixlive - thanks - and done!

PR #234, opened 2025-03-07
fix(anthropic): tool choice must be object

mattmcdev:
For Anthropic all tool choices (not just named tools) must be returned in an object with a `type` key (https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#example-simple-tool-definition).

This PR fixes that, and also adds a mapping for `ToolChoice::None`


sixlive (approved):
Thanks!

PR #235, opened 2025-03-07
feat: revert required tool calling after n steps

mattmcdev:
Opening this PR to replace [this one](https://github.com/echolabsdev/prism/pull/178).

This proposes adding a new option when calling `withToolChoice` to automatically revert the specific choice back to `auto`.

When combined with setting `maxSteps` this lets you require a model use a tool n times and then return a response.

The default value is `null` so should be a non-breaking change.

Can update docs on how and when to use this if accepted.

### DIFF
--- a/src/Concerns/ConfiguresTools.php
+++ b/src/Concerns/ConfiguresTools.php
@@ -11,11 +11,15 @@ trait ConfiguresTools
 {
     protected string|ToolChoice|null $toolChoice = null;
 
-    public function withToolChoice(string|ToolChoice|Tool $toolChoice): self
+    protected ?int $toolChoiceAutoAfterSteps = null;
+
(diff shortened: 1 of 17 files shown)


mattmcdev:
@sixlive This might be an OpenAI specific issue - not sure any other providers have `required` as a tool choice.

The (reproducible) problem, is if you set `withToolChoice(ToolChoice::Auto)` and `withMaxSteps`, you're going to loop until `maxSteps` is hit, and just call the tool on each step. You'll never get any response back.

Maybe instead of a global `toolChoiceAutoAfterSteps`, I should implement this as a `providerOption` for OpenAI specifically ? Or make it a default for OpenAI that after the first use of 'required' it reverts back to 'auto' ?

Let me know what you think, would be good to get this issue solved.

sixlive:
I'll revisit this in the future, going to leave the current functionality for now

PR #236, opened 2025-03-07
feat(openai): support required tool calling

mattmcdev:
This PR just adds a valid mapping for `ToolChoice::Any` with OpenAI and updates the docs to remove it as a limitation.


sixlive:
Can we add a test to validate that this works as expected?

mattmcdev:
Done

PR #242, opened 2025-03-08
feat(mistral): rate limit exception and meta

ChrisB-TL:
## Description
Implements the rate limit exception and meta for Mistral.

Note that the exception side of this is done a little bit on guess work (that a rate limit hit returns 429 with the same headers as a normal response), as its undocumented and figured out from experimentation... Confident in the successful request path (as far as can be without Mistral providing docs!).

Note: I have not updated the docs page as I am doing a couple of these today and don't want to cause a conflict. Will circle back after PRs are merged to update.


sixlive (approved):

sixlive:
Looks like we have a conflict that needs resolving before merge

ChrisB-TL:
@sixlive thanks - should now be fixed!

PR #247, opened 2025-03-09
fix: misleading tool return type exception

kauffinger:
## Description

When a tool returns something else than string, currently `Invalid parameters for tool : %s` is thrown. This message misled me for a moment. The only way to find out the correct reason is to check the previous exception.

As per type definition, a string must be returned from the tool, I think a good place to check for that is right after getting the tool's value. Also, this way, the exception throwing is co-located with the reason it is needed.

This PR adds a more user-friendly exception message (`Invalid return type for tool : %s. Tools must return string.`) and a sentence about tool return types in the tool documentation.

## Breaking Changes

Shouldn't be breaking as tools without `string` return type cannot be used anyways. The only way this could be breaking in my head is if a package consumer let's their consumer/user dynamically define the return type & checks for the current error message while catching.

### DIFF
--- a/docs/core-concepts/tools-function-calling.md
+++ b/docs/core-concepts/tools-function-calling.md
@@ -62,6 +62,8 @@ $searchTool = Tool::as('search')
     });
 ```
 
+Tools can take a variety of parameters, but must always return a string.
+
 ## Parameter Definition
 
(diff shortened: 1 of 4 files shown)


sixlive (approved):
Love this! Thanks!

PR #253, opened 2025-03-10
bug(openai + groq): fix null rate limit reset

ChrisB-TL:
## Description

Avoids an exception for openai and groq rate limits where the resets header is null.


sixlive (approved):
Good catch!

PR #256, opened 2025-03-12
feat(openai): Responses API

mattmcdev:
This change migrates OpenAI from `/chat/completions` to [the new Responses API](https://platform.openai.com/docs/api-reference/responses).

I believe it is (almost) non-breaking, and the only newly introduced features are support for the new [`previous_response_id` argument](https://platform.openai.com/docs/api-reference/responses/create#responses-create-previous_response_id) and [`truncation` argument](https://platform.openai.com/docs/api-reference/responses/create#responses-create-truncation) when passed in `providerMeta`.

The only breaking change - the `callId` property of `ToolCall` I've renamed to `id` because OpenAI now requires both an `id` and `call_id` for tool calling, so it was that or something like `callCallId`. To avoid making this breaking, perhaps setting `callId` equal to `id` when it's not provided ?

I also changed the `providerMeta` function in `HasProviderMeta` to return null when the `valuePath` can't be found. I assume it's meant to return the entire object when `valuePath` is `null`, which it still does with this change. Before it also returned the entire object when `valuePath` was set but the path returned `null` - I couldn't see anywhere that seemed to want that behaviour - maybe it was a bug ? 

Why this change ? As well as supporting the `previous_response_id` to help easily keep context in long threads, the Responses API is also their [recommended option for new users](https://platform.openai.com/docs/guides/responses-vs-chat-completions).

⚠️ Tests are currently failing because I've not updated the streaming fixtures - if you're happy with this proposed change I'll get those updated to the new format and document `previous_response_id`


sixlive:
I'll take a closer look at this later today. I haven't done any research on the responses API yet.

mattmcdev:
> I'll take a closer look at this later today. I haven't done any research on the responses API yet.

Thank you - I think there's other provider metadata that could be supported, but wanted to keep this mainly focused on switching to the Responses API and then build upon that. Just added `previous_response_id` because it's an immediate fix for a chat thread length issue I've been having !

sixlive:
Sorry for the MASSIVE delay on this. I did the research, I'm down to move over to the responses API

rzv-me:
hi, I am also interested on this, as you can't have reasoning without Response API

rzv-me:
> Should have some time to update the PR next week

Anything I can help with?

mattmcdev:
> > Should have some time to update the PR next week
> 
> Anything I can help with?

Main issues are it needs rebasing against main, and the tests needs some TLC. 

Separately, to work with `o4-mini` I had to workaround [an undocumented requirement to send back an empty reasoning input](https://community.openai.com/t/responses-api-invalid-request-error-function-call-was-provided-without-its-required-reasoning-item) when using function calls - threw [this](https://github.com/mattmcdev/prism/commit/537b94fc172b77965ee4c955096a9b139f926152) together in the internal build  I use, but needs some thinking about.

PR #260, opened 2025-03-13
feat(openai): retry after header

mattmcdev:
This matches the `retryAfter` value of an OpenAI 429 rate limited response using the [same header as their Python library](https://github.com/openai/openai-python/blob/e36956673d9049713c91bca6ce7aebe58638f483/src/openai/_base_client.py#L621)


ChrisB-TL:
Good find! Can we add this to the existing tests?

mattmcdev:
Yep - updated the rate limit test

sixlive (approved):
Thanks!

PR #261, opened 2025-03-13
feat(openai): set the name of unmatched OpenAI rate limit errors

mattmcdev:
Handles OpenAI rate-limit errors (e.g. `insufficient_quota`) where no headers are set and the error is provided in the response.

### DIFF
--- a/src/Providers/OpenAI/Concerns/ProcessesRateLimits.php
+++ b/src/Providers/OpenAI/Concerns/ProcessesRateLimits.php
@@ -26,6 +26,19 @@ trait ProcessesRateLimits
             $rateLimits[$limitName][$fieldName] = $headerValues[0];
         }
 
+        if ($rateLimits === []) {
+            $error = data_get($response->json(), 'error');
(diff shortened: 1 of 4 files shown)


ChrisB-TL:
Frustratingly, I don't think this is documented anywhere! I take it you have come across this with a real response?

Assuming tricky to test this with a real request, but can we add a test for this with manually built response and `FixtureResponse`? Would need to add a status code prop to the response sequence method.

mattmcdev:
Yeah real response - have created a fixture for it

ChrisB-TL:
Great! The only ones I've "road tested in prod" are Anthropic so good to have some real test cases..!

mattmcdev:
So the various `'x-ratelimit-` headers in`ProcessesRateLimits` - are those something you've seen documented ? I couldn't see any reference to them in their Python or Node libraries.

There is logic to decide whether a request is retryable which could be a useful addition to the `providerResponseError`.

sixlive (approved):
Thanks for this!

PR #262, opened 2025-03-14
feat(mistral): Support mistral documents in chat conversation

LarsWiegers:
## Description
1. Updates the Anthropic and Gemini providers to check if an url document is passed and throw an exception if so. The support for these providers for url based documents can be added in another pr.
2. Adds an fromUrl method constructor to the Document value object so that users can create documents from url.
3. Updates the Mistral provider to allow documents based on urls to be passed in.
4. Added tests for document value object
5. Added tests for document handeling in Mistral provider

## Breaking Changes

1. Updates the Mistral ToolChoiceMap to allow the user not to pass in a tool. This would previously throw an exception which did make sense to me as the api allows no toolchoice to be set. If this is wrong please let me know.

PS: The key that was exposed has been deleted.


LarsWiegers:
Let me know if I should change anything, this was a great learning experience :)

sixlive (approved):
Thanks for this!!

PR #263, opened 2025-03-14
Refactor(tests): FixtureResponse Improvements

sixlive:
## Description

This PR refactors the `FixtureResponse` test helper to automatically record the fixture if the fixture file is not present. This makes it MUCH easier to reason about. It also makes it easier to re-record fixtures since we'll want to do this somewhat more regularly. 

This PR also re-records all of the OpenAI fixtures and refactors the OpenAI tests to be more resilient to updated fixtures.


ChrisB-TL:
Great idea. Can we add a property to the sequence method to force re-record?

Just a nicer workflow when debugging trying to get the "right" response - to not have to find and delete the file each time.

sixlive:
Love that idea!

PR #265, opened 2025-03-15
feat(gemini): text search grounding

ChrisB-TL:
## Description

This PR:
- allows you to enable [Google search grounding](https://ai.google.dev/gemini-api/docs/grounding?lang=rest) on text requests
- returns the data you need to build up footnotes and comply with Google's display requirements in `additionalContent`.

Gemini does not support:
- use of search grounding with other tools, without using the "Live" (websocket) API (so Prism will throw an exception if you try).
- search grounding for structured.


sixlive (approved):
Sick!

PR #270, opened 2025-03-17
docs(stream): fix stream docs

uchm4n:
## Description
Fix stream docs by updating old syntax to the new one.


sixlive (approved):
Oops! Totally missed this, great catch, thanks!!

Issue #282, opened 2025-03-27
"Array to string conversion" error when using wrong parameters for Mistral AI call


laundy:
I'm using prism `0.56.0` in a fresh Laravel `12.3.0` installation. When building a very basic example (see below), I get a generic `Array to string conversion` exception which points to the line `asText()`.

```php
$response = Prism::text()
            ->using( Provider::Mistral, 'ministral-8b-latest' )
            ->withSystemPrompt( 'Act as a service agent.' )
            ->withPrompt('Hello, how can I do this or that?')
            ->asText();
```

We [see](https://github.com/prism-php/prism/blob/bb662ec2cf9187885bdf301ac83fa290d8e9ee08/src/Providers/Mistral/Concerns/ValidatesResponse.php#L27-L33) see that you are using `data_get($data, 'message', 'unknown')` for composing the exception message, but Mistral AI's API can also contain an array as message as this `dd` output shows:

```
array:5 [▼ // vendor/prism-php/prism/src/Providers/Mistral/Concerns/ValidatesResponse.php:26
  "object" => "error"
  "message" => array:1 [▼
    "detail" => array:1 [▼
      0 => array:4 [▼
        "type" => "float_type"
        "loc" => array:2 [▼
          0 => "body"
          1 => "top_p"
        ]
        "msg" => "Input should be a valid number"
        "input" => null
      ]
    ]
  ]
  "type" => "invalid_request_error"
  "param" => null
  "code" => null
]
```

Also, the Mistral AI docs seem to be wrong about this as well: https://docs.mistral.ai/api/#tag/chat/operation/chat_completion_v1_chat_completions_post as it tells us that `message` must be a string. For this I've created a separate issue at their repository (https://github.com/mistralai/platform-docs-public/issues/212).

In addition, Mistral AI's API is not accepting sending `"top_p": null,` with a request, hence it throws the error mentioned above and we need to call `usingTopP( <float> )` when building a prompt. Casting `null` as `float` is `0.0` and Mistral AI also does not like it: `top_p must be in (0, 1], got 0.0.`

sixlive:
Merged #283

PR #283, opened 2025-03-27
fix(mistral): error message handling

laundy:
One approach for solving issue #282.


ChrisB-TL:
On top_p, this shouldn't be necessary, as `array_filter` removes null values from the payload. It looks like its just missing from the text handler. The fix for this is to wrap the payload in `array_filter` as we do in the structured handler.

The error fix looks good.

laundy:
Thanks for your review and the hint! I've moved all optional parameters to the `array_filter` in `Text`.

sixlive (approved):
Awesome, thank you so much!

PR #286, opened 2025-03-29
bug(openai): Fixed Text handler response

EthanBarlo:
## Description
I was testing out the `PrismServer` feature and found that OpenAi models were not returning a response.

I forked to debug and found this line that was missed during a refactor.

Updating it fixed my issue and `PrismServer` now returns the correct response.


sixlive (approved):
Oh dang! Really good catch! Thanks!

Issue #110, opened 2024-12-28
Options feature


mj-azimi:
Please add the options value in ollama.

https://github.com/ollama/ollama/blob/main/docs/api.md

curl http://localhost:11434/api/generate -d '{
  "model": "llama3.2",
  "prompt": "Why is the sky blue?",
  "stream": false,
  "options": {
    "num_keep": 5,
    "seed": 42,
    "num_predict": 100,
    "top_k": 20,
    "top_p": 0.9,
    "min_p": 0.0,
    "typical_p": 0.7,
    "repeat_last_n": 33,
    "temperature": 0.8,
    "repeat_penalty": 1.2,
    "presence_penalty": 1.5,
    "frequency_penalty": 1.0,
    "mirostat": 1,
    "mirostat_tau": 0.8,
    "mirostat_eta": 0.6,
    "penalize_newline": true,
    "stop": ["\n", "user:"],
    "numa": false,
    "num_ctx": 1024,
    "num_batch": 2,
    "num_gpu": 1,
    "main_gpu": 0,
    "low_vram": false,
    "vocab_only": false,
    "use_mmap": true,
    "use_mlock": false,
    "num_thread": 8
  }
}'

sixlive:
We need to address #130 before we can move forward with this.

sixlive:
Support for this was merged!

PR #288, opened 2025-03-31
feat(model-options): add HasModelOptions trait and integrate model op…

uchm4n:
## Description
As of now, Prism doesn't provide a way for library users to specify additional options to model providers beyond the standard `temperature, topP, and maxTokens (num_predict)` parameters. Many providers offer a rich set of configuration options that our users currently cannot access.

To make Prism more flexible, I've introduced a new `HasModelOptions` trait to provide custom configuration options to all model providers via a same, fluent API:
```php
 Prism::text()
    ->using(Provider::Anthropic, 'claude-3-sonnet')
    ->withSystemPrompt(view('prompts.system'))
    ->withPrompt('Explain quantum computing to a 5-year-old.')
    ->withClientOptions(['timeout' => 30])
    ->withOptions([ // TODO: Planed to add
	    'temperature' => 0.9,
	    'num_predict' => 1000,
	    'top_p' => 0.9,
            'top_k' => 40,
            'seed': 42
    ])
    ->asText();

```
This implementation allows users to leverage provider-specific options beyond our standard parameters, making Prism more adaptable to various use cases without requiring core changes for each new parameter.

One consideration in the API design is the potential confusion between `Prism::text()->withOptions()` and Laravel's HTTP facade `Http::withOptions()`. Despite the naming similarity, I felt this approach best matches our existing fluent API pattern.

Currently I implemented this feature for the Ollama provider as a proof of concept. If feedback is positive, I'll extend this functionality to other providers in this PR. (PR solves issue: #110 )
Looking forward to your feedback @sixlive

### DIFF
--- a/src/Concerns/HasModelOptions.php
+++ b/src/Concerns/HasModelOptions.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Prism\Prism\Concerns;
+
+use Prism\Prism\Text\PendingRequest;
+
+trait HasModelOptions
+{
+    /** @var array<string, array<string, mixed>> */
+    protected array $modelOptions = [];
+
+    /**
+     * @return HasModelOptions|\Prism\Prism\Structured\PendingRequest|PendingRequest
+     */
+    public function withOptions(array $options): self
+    {
+        $this->modelOptions = array_merge($this->modelOptions, $options);
+
+        return $this;
+    }
+}
--- a/src/Concerns/HasModelOptions.php
+++ b/src/Concerns/HasModelOptions.php
@@ -2,15 +2,13 @@
 
 namespace Prism\Prism\Concerns;
 
-use Prism\Prism\Text\PendingRequest;
-
 trait HasModelOptions
 {
-    /** @var array<string, array<string, mixed>> */
+    /** @var array<string, mixed> */
     protected array $modelOptions = [];
 
     /**
-     * @return HasModelOptions|\Prism\Prism\Structured\PendingRequest|PendingRequest
+     * @param  array<string, mixed>  $options
      */
     public function withOptions(array $options): self
     {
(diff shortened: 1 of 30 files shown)


ChrisB-TL:
I like this, but we will need to make clear in the docs the different use case between this and `withProviderMeta`. I.e.

- This is for making option changes directly that don't need a package change (I guess "unofficially supported by Prism"); and
- providerMeta is for making option changes in an officially supported way.

benbjurstrom:
Would love to see this merged. Ollama limits all models to a context window size of 2048 tokens by default[^1]. This will let us pass num_ctx to customize the size of the window.

[^1]:https://github.com/ollama/ollama/blob/main/docs/faq.md#how-can-i-specify-the-context-window-size

uchm4n:
Yes, you’re absolutely right, @ChrisB-TL . I’ll make sure to document withProviderMeta and `withOptions` clearly.

I want to highlight that `withOptions` is meant for configuring options like `temperature` and `top_p`, but some providers offer many additional settings. This approach ensures users have a way to configure those options flexibly. Also, if a provider introduces new options in the future, withOptions would be the only way to support them without requiring a package update.

For now, I’ve added this for each provider, but I still need to add tests and documentation accordingly. If you’d like to tweak it, you’re more than welcome to do so.

sixlive:
So, I guess, I was thinking about putting Ollama options in provider meta, I think thats kind of what I envisioned it for... provider specific options and flags

uchm4n:
I guess having options in to one single place is making sense.
Also `withProviderMeta` is not that particularly descriptive name in my opinion.

sixlive:
I think we could rename it from `withProviderMeta` to `withProviderOptions`?

uchm4n:
That would be much more readable and intuitive name. Than I guess we can close this PR

PR #290, opened 2025-04-01
feat(openai): add support for file inputs

Gertiozuni:
## Description

This PR adds support for file inputs/documents:

https://platform.openai.com/docs/guides/pdf-files?api-mode=chat

### DIFF
--- a/docs/input-modalities/documents.md
+++ b/docs/input-modalities/documents.md
@@ -41,6 +41,7 @@ use Prism\Enums\Provider;
 use Prism\Prism\Prism;
 use Prism\Prism\ValueObjects\Messages\UserMessage;
 use Prism\Prism\ValueObjects\Messages\Support\Document;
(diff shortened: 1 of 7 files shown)


ChrisB-TL:
Thanks for this! 

I am not sure about the support for file IDs. They are OpenAI specific, and we don't offer a way to upload documents via Prism, so it may throw up a lot of issues/questions.

If we do leave them in, could we move to a OpenAI specific `OpenAIFile` value object - to be added to a message in the same way as `Document` - which should avoid confusion a little?

ianhawes:
PDF files can be added inline via Base64 as documented here:
https://platform.openai.com/docs/guides/pdf-files?api-mode=chat#base64-encoded-files

sixlive (approved):
Thanks for this!

elnurvl:
The documentation still not updated to advertise documents support for OpenAI. Is there a catch?

PR #291, opened 2025-04-02
Update testing.md

kyranb:
Properties are now protected and need to be accessed through the public method.

## Description

Properties are now protected and need to be accessed through the public method. So the docs have been updated to reflect that.


sixlive (approved):
Awesome thanks!!

PR #298, opened 2025-04-06
feat(deepseek): support api url configuration

JaguarJack:
## Description
The deepseek API URL in the current project is fixed and official. Currently, there are many transfer platforms, such as openrouter, which also support it. So you need to change the deepseek API URL to a configurable file.


sixlive (approved):
This is great! Thanks!

pushpak1300 (approved):

PR #303, opened 2025-04-09
feat(server): system messages

edvinbrobeck:
## Description
This PR simply adds `SystemMessage` to the mapping in `PrismChatController`.
I was using Open WebUI with a knowledge base, and the "RAG context" is added with `'role' => 'system'`.


sixlive (approved):
AWESOME!! Thank you! I meant to do this at somepoint but lost track of it, thanks!

PR #308, opened 2025-04-11
chore(docs): align code example with ollama provider

kauffinger:
While working through the providers I found this code snippet. Thought I'd align it with the to not confuse beginners.


sixlive (approved):
Good catch! Thanks!

PR #311, opened 2025-04-13
fix(gemini): image from url fix

uchm4n:
## Description
This PR fixes image analysis from URLs for `Gemini` models. Initially, this feature was implemented for `Anthropic` but was never tested on `Gemini`. `Gemini`, however, supports images from URLs as long as the content is base64 encoded.

To identify which provider is being used, I registered `ProviderEnum` in the IoC container to retrieve it later. (This approach can be reused in the future if we need to check provider instances per feature.)

Additionally, I fixed the related tests.


ChrisB-TL:
I am currently working on a rationalisation of how we handle images and documents to handle different providers' levels of support neatly. I'll work this in?

sixlive:
Yeah, thanks @ChrisB-TL

ChrisB-TL:
@uchm4n see draft PR #321 - thoughts welcome!

sixlive:
I'm gonna go ahead and close this for now in favor of what @ChrisB-TL is working on. Lets move the conversation over this #321

Issue #205, opened 2025-02-23
Feature: Unhandled finish reasons


ChrisB-TL:
Post refactor, I think we should add a specific exception for unhandled finish reasons, for folks to hook into. 

I'd envisage it being thrown at the bottom of the handle method on provider handlers and taking the provider, the mapped finished reason and the "raw" provider finish reason.

```php

return match ($finishReason) {
    FinishReason::ToolCalls => $this->handleToolCalls($data, $request),
    FinishReason::Stop => $this->handleStop($data, $request),
    default => throw new PrismUnhandledFinishReasonException(Provider::Gemini, $finishReason, $finishReasonRaw),
};
```

Am happy to pick this up if you agree @sixlive, just leaving this here as a reminder to circle back to it.

ChrisB-TL:
I'd actually handle length specifically the same as stop - as IMO its only a quasi-error state. I think we do this for some providers already (Anthropic + Gemini, I think).

@sixlive - perhaps we can mark this issue as a good first PR?

LasseRafn:
or perhaps if handling length is too troublesome , can include the finish reason / response in the exception, so that we can implement our own handling?

PR #313, opened 2025-04-16
Add OpenAI response to PrismException

aglipanci:
## Description
Adds data attribute to the PrismException exception to be able to pass more data based on the provider to allow more control on the user end. The most obvious use case is when OpenAI returns a finish_reason: length, which is valuable information to have on the exception.

## Breaking Changes
None

### DIFF
--- a/src/Exceptions/PrismException.php
+++ b/src/Exceptions/PrismException.php
@@ -10,6 +10,27 @@ use Throwable;
 
 class PrismException extends Exception
 {
+    public function __construct(
+        string $message = '',
+        int $code = 0,
+        ?Throwable $previous = null,
+        protected ?array $data = null
+    ) {
+        parent::__construct($message, $code, $previous);
+    }
+
+    public function getData(): ?array
+    {
+        return $this->data;
+    }
+
+    protected function setData(?array $data): self
+    {
+        $this->data = $data;
+
+        return $this;
+    }
+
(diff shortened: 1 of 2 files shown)


ChrisB-TL:
> > See also #205 re: the specific example given.
> 
> My approach here is to at least include the response in the exception, so that we can have some control. I don't really know the difference between OpenAI and other providers in terms of response format or errors but this is not a breaking change and it offers space for adding more data to other providers in the future.

My preference here is per #205, with a specific unhandled finish reason exception than can be caught rather than the generic `PrismException`. The example in that issue would expose this same data.

Generally, where more context is needed on an exception we have implemented a specific exception with strong types - e.g. for rate limits.

aglipanci:
@ChrisB-TL, thanks for taking the time to get back at me about this :)

I agree with you, and I personally would prefer the specific exception as well. However, since that seems like a lot of work (I would be happy to help, of course, but I don't have much experience with all the providers), I thought at least some visibility in the data would be helpful. 

You can close this PR if you see fit; it's fine with me.

ChrisB-TL:
Happy to support/review if you'd like to have a stab at it.

All of the structured/text providers follow the same pattern (`match` block at the end of the `handle` method), so should be reasonably straight forward. Stream works differently, so we could leave that for now.

PR #321, opened 2025-04-23
refactor: better media handling

ChrisB-TL:
## Description

Our `Document` and `Image` value objects are becoming slightly brittle when used between providers. This PR seeks to address that and make it easier to add different modalities in the future (e.g. Video, Audio).

- This PR refactors both to extend a `Media` class.
- The `Media` class is completely provider agnostic, and gracefully handles switching between mediums (with some limitations).
- Adds Laravel `Storage` support.
- Extensions of the new `ProviderMediaMapper` abstract class are responsible for defining what a provider can support, handling exceptions and converting to a suitable payload. Each provider will extend this class for `Image` and `Document`.
- Each providers `MessageMapper` will defer to the respective `ProviderMediaMapper` when converting an `Image` or `Document` to a payload.

Re: switching between mediums, in summary:
- Where a provider only supports URLs: if you provide a file path, contents, base64 or chunks, for security reasons Prism does not create a URL for you and your request will fail. The user would need to create their own URL (which could be single use). We could do this for the dev, but for security conscious use cases (like my own) this would make me nervous. 
- Where a provider does not support URLs: the `base64()` and `rawContent()` methods fetch the URL and return base64/raw.
- Where you have base64 or rawContent: Prism will switch between them depending on what the provider accepts.

So far I have implemented the shared logic and updated the Anthropic provider which covers every medium for a very full example. If we are happy with the approach, I will implement for remaining providers as part of this PR. Failing tests are due to having not done this yet.

## Breaking Changes
- `Image` and `Document` no longer have constructors to ensure object consistency. Use the `::from` methods instead.
- `Image` and `Document` properties have changed, are now protected, and are now accessed via accessor methods to allow some magic. This should only affect custom providers that use `Image` and `Document`.
- Anthropic only: the static `from` methods on `Document` value objects no longer accept context, as this is unique to Anthropic. Use `->withProviderMeta(Provider::Anthropic, ['context' => 'here is your context'])` instead.
- Depreciated `Image` and `Document` `::fromPath()` in favour of `::fromLocalPath()` to avoid confusion with `::fromStoragePath()`.


src/ValueObjects/Messages/Support/Media.php
```diff
@@ -0,0 +1,220 @@
...
+            throw new InvalidArgumentException("$path does not exist on the '$disk' disk");
+        }
+
+        $content = $disk->get($path);
+
+        if ($content === '' || $content === '0' || $content === false) {
+            throw new InvalidArgumentException("$path on the '$disk' disk is empty.");
+        }
+
+        $mimeType = (new finfo(FILEINFO_MIME_TYPE))->buffer($content);
```
uchm4n, line 77: Laravel Storage has mimeType function, I think it would be appropriate to use it here. 
```suggestion
        $mimeType = $disk->mimeType($path);
```

src/ValueObjects/Messages/Support/Media.php
```diff
@@ -0,0 +1,220 @@
...
+
+    public static function fromLocalPath(string $path): static
+    {
+        if (! is_file($path)) {
+            throw new InvalidArgumentException("$path is not a file");
+        }
+
+        $content = file_get_contents($path);
+
+        if ($content === '' || $content === '0' || $content === false) {
```
uchm4n, line 44: content could be one of these: - `''`, `'0'`, `0`, `0.0`, `null`, `false`, `[]`
```suggestion
        if (empty($content)) {
```

src/ValueObjects/Messages/Support/Media.php
```diff
@@ -0,0 +1,220 @@
...
+    {
+        $disk = Storage::disk($disk);
+
+        if ($disk->exists($path) === false) {
+            throw new InvalidArgumentException("$path does not exist on the '$disk' disk");
+        }
+
+        $content = $disk->get($path);
+
+        if ($content === '' || $content === '0' || $content === false) {
```
uchm4n, line 73:
```suggestion
        if (empty($content)) {
```

src/ValueObjects/Messages/Support/Media.php
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        return $this->mimeType;
+    }
+
+    public function fetchUrlContent(): void
+    {
+        $content = Http::get($this->url)->body();
+
+        if ($content === '' || $content === '0' || $content === false) {
```
uchm4n, line 208:
```suggestion
        if (empty($content)) {
```

ChrisB-TL:
@uchm4n Good catch re: `fromRawContent` - this is an error in the docs. Both signatures currently require a mimeType, though we could/should remove this if we keep the finfo dependency.

@sixlive - are you happy with the finfo dependency? If so, I'll add it to composer, and if not, I'll find another solution. I used it for a few reasons:
- The Laravel Storage driver doesn't have a `mimeType` method (though I could have sworn it did!)
- When loading from a URL, it avoids relying on headers.
- It allows loading from strings/base64 without specifying a mimeType (DX only)

AFAIK its bundled by default other than on windows. Laravel currently requires it, as thephpleague/flysystem does.

src/ValueObjects/Messages/Support/Media.php
uchm4n: https://laravel.com/docs/11.x/filesystem#file-metadata -> `$mime = Storage::mimeType('file.jpg');`

src/ValueObjects/Messages/Support/Media.php
ChrisB-TL: Ah, I thought it did but my IDE complained so thought I was going mad. Tested, and this does work, despite VSCode's red squiggle...

src/ValueObjects/Messages/Support/Media.php
kauffinger: The extended if is probably from the rector that transforms empty() into those three assertions.

laravel's blank() helper will do the job

uchm4n:
```
blank('');           // true
blank('0');          // false -> here if content has 0 it will be considered not empty
blank(null);         // true
blank([]);           // true
blank(collect([]));  // true
blank(0);            // false -> here too
```
If this is ok than I guess to use `blank()` make sense here

kauffinger: Yeah sorry, should have noted the different behavior - you can think of blank as a "strict" empty

ChrisB-TL: So many versions of empty! Prism doesn't actually have that rector, but I assumed it did from looking at the existing code. So we can just use `empty(trim($content))`

kauffinger: Hahah well, that's awkward, I assumed on that basis, too. I like your suggestion, the `blank` helper always looks a little crowded to me. Less useless checks & no dependency 🎉

ChrisB-TL: I take this back. Rector wants one of these (depending on situation) instead of my suggestion 😆 

```php
in_array(trim($content), ['', '0'], true)
```

```php
in_array(trim($content ?? ''), ['', '0'], true)
```

I'm going to comply with it!

uchm4n:
@ChrisB-TL, I hope you're doing well! Could you let us know if there's anything needed to finalize this PR? Happy to assist if there’s anything holding it up.

ChrisB-TL:
@uchm4n - sorry for the slow reply! I'd just like @sixlive to take a look and check he is happy with approach before I finish off the other providers.

sixlive:
@ChrisB-TL I'm down for heading in this direction

ChrisB-TL:
@sixlive awesome - have implemented the remaining providers so this should be good for final review.

PR #323, opened 2025-04-23
fix: providerMeta returns null if valuePath is not set

ChrisB-TL:
## Description
I have been meaning to do this for a while, but in my head it was a bigger change than it actually was...!

`->providerMeta()` currently returns the full providerMeta for the specified provider (or an empty array) if the valuePath is not set. This is clunky and requires workarounds to handle null values.

This PR changes `->providerMeta()`  to:
- Return early if the valuePath is null (therefore not affecting behaviour if your intent is to grab the whole array).
- Return null if the valuePath is not set.

I've also been through and cleaned up where we have worked around this internally.

## Breaking Changes
- `->providerMeta()` now returns null if the valuePath is not set, instead of returning the full providerMeta array for the provider. This _should_ only affect custom providers.

### DIFF
--- a/src/Concerns/HasProviderMeta.php
+++ b/src/Concerns/HasProviderMeta.php
@@ -27,6 +27,10 @@ trait HasProviderMeta
             []
         );
 
-        return data_get($providerMeta, $valuePath, $providerMeta);
+        if ($valuePath === null) {
+            return $providerMeta;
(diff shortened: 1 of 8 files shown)


sixlive (approved):
YESSS! Thanks!

PR #324, opened 2025-04-23
feat(docs): add local search to documentation

uchm4n:
## Description
This is a tiny change but supper helpful. 
`VitePress` has this [local search](https://vitepress.dev/reference/default-theme-search#local-search) feature, which is in-browser index offline search. It's not that advanced but in our case it's more than enough.


sixlive (approved):
YES!! Awesome!! Thank you so much for this, its been on my mind for a minute but just couldn't get to it.

PR #331, opened 2025-04-26
feat: add providerConfig argument to using

ChrisB-TL:
## Description

Adds a providerConfig argument to `using()`, e.g.:

```php
Prism::text()
    ->using(Provider::OpenAI, 'gpt-4', [
        'url' => 'new url'
    ]);
```
You can do this already with `usingProviderConfig()`, but this is slightly nicer DX and avoids re-resolving the provider.


sixlive (approved):
Sick, I dig this, thanks!

PR #332, opened 2025-04-26
Expand OpenAI tool choice mapping to support 'Any' and 'None' options

mirceasoaica:
Added mappings for ToolChoice::Any ('required') and ToolChoice::None ('none'). Keeps existing behavior for Auto, null, and invalid values.


sixlive (approved):
Awesome, thanks!

PR #335, opened 2025-04-26
feat: add whenProvider for provider interop

sixlive:
## Description

This PR introduces `whenProvider` as a way to provide a path to easier provider interop.


ChrisB-TL:
Love it!

kauffinger:
Yup, I like that, the invokable class also makes sense for more complex scenarios - let's say configuring based on a provider's model.

Great ship! 🚀

PR #336, opened 2025-04-26
feat(mistral): add stream support for Mistral provider

uchm4n:
## Description
Add stream support for Mistral provider


sixlive (approved):
Sick! Thanks!!

PR #338, opened 2025-04-27
chore: freeze time for tests

kauffinger:
## Description
The PrismChatController tests would fail sometimes due to timing, as the time was frozen after the responses were generated. Thus, they would usually work until a runner is too slow. To make sure they don't, it is now frozen at start of the test.

## Breaking Changes
none


sixlive (approved):
Thanks for this!!!

PR #344, opened 2025-04-29
feat(gemini): Add support for Gemini 2.5 Thinking Budget

btumbleson:
## Description
In the preview Gemini 2.5 models, Google introduced "thinking", and is enabled by default on these models. Without this feature, using the new models will always include thinking, which is billed at a higher rate and not always desirable. This PR enables you to configure a `thinkingBudget`, which Gemini views a max amount of tokens that may be used for thinking, or it can be set to 0 to disable it altogether.

This PR:

- Introduces a new `thoughtTokens` to the `Usage` class
- Leverages the `withProviderOptions()` method to inject a `thinkingBudget` when passed as a key, value pair
- Includes two tests that check for `thoughtTokens` in the `Usage` output
- Updates the documentation for Gemini to include this functionality

## Usage

```
$response = Prism::text()
			->using(Provider::Gemini, 'gemini-2.5-flash-preview')
			->withPrompt('Explain the concept of Occam\'s Razor and provide a simple, everyday example.')
			->withProviderOptions(['thinkingBudget' => 500])
			->asText();
```

You can set `thinkingBudget` to `0` to disable thinking altogether. Note this also works with structured.

## Potential Discussion
Gemini has made the choice that thinking is on by default, so this PR follows that convention. As in, if you do not specify a `thinkingBudget` of `0`, then thinking will be enabled. An alternative approach could assume that unless you include a `thinkingBudget`, that Prism can set this value to 0. For this to work effectively you would need to (re)introduce a `ThinkingModeResolver`, at least in the near term, as I expect all models will eventually gravitate towards this thinking convention.

### DIFF
--- a/src/Providers/Gemini/Handlers/Stream.php
+++ b/src/Providers/Gemini/Handlers/Stream.php
@@ -242,11 +242,11 @@ class Stream
                             'temperature' => $request->temperature(),
                             'topP' => $request->topP(),
                             'maxOutputTokens' => $request->maxTokens(),
-							'thinkingConfig' => array_filter([
-								'thinkingBudget' => array_key_exists('thinkingBudget', $providerOptions)
-									? $providerOptions['thinkingBudget']
-									: null,
-							], fn($v) => $v !== null),
+                            'thinkingConfig' => array_filter([
+                                'thinkingBudget' => array_key_exists('thinkingBudget', $providerOptions)
+                                    ? $providerOptions['thinkingBudget']
+                                    : null,
+                            ], fn ($v) => $v !== null),
                         ]),
                         'tools' => $tools !== [] ? $tools : null,
                         'tool_config' => $request->toolChoice() ? ToolChoiceMap::map($request->toolChoice()) : null,
--- a/src/Providers/Gemini/Handlers/Stream.php
+++ b/src/Providers/Gemini/Handlers/Stream.php
@@ -243,8 +243,10 @@ class Stream
                             'topP' => $request->topP(),
                             'maxOutputTokens' => $request->maxTokens(),
 							'thinkingConfig' => array_filter([
-								'thinkingBudget' => $providerOptions['thinkingBudget'] ?? null,
-							]),
+								'thinkingBudget' => array_key_exists('thinkingBudget', $providerOptions)
+									? $providerOptions['thinkingBudget']
+									: null,
+							], fn($v) => $v !== null),
                         ]),
                         'tools' => $tools !== [] ? $tools : null,
                         'tool_config' => $request->toolChoice() ? ToolChoiceMap::map($request->toolChoice()) : null,
--- a/src/Providers/Gemini/Handlers/Stream.php
+++ b/src/Providers/Gemini/Handlers/Stream.php
@@ -242,6 +242,9 @@ class Stream
                             'temperature' => $request->temperature(),
                             'topP' => $request->topP(),
                             'maxOutputTokens' => $request->maxTokens(),
+							'thinkingConfig' => array_filter([
+								'thinkingBudget' => $providerOptions['thinkingBudget'] ?? null,
+							]),
                         ]),
                         'tools' => $tools !== [] ? $tools : null,
                         'tool_config' => $request->toolChoice() ? ToolChoiceMap::map($request->toolChoice()) : null,
--- a/src/Providers/Gemini/Handlers/Structured.php
+++ b/src/Providers/Gemini/Handlers/Structured.php
@@ -106,8 +106,9 @@ class Structured
                     data_get($data, 'candidates.0.finishReason'),
                 ),
                 usage: new Usage(
-                    data_get($data, 'usageMetadata.promptTokenCount', 0),
-                    data_get($data, 'usageMetadata.candidatesTokenCount', 0)
+                    promptTokens: data_get($data, 'usageMetadata.promptTokenCount', 0),
+					completionTokens: data_get($data, 'usageMetadata.candidatesTokenCount', 0),
+					thoughtTokens: data_get($data, 'usageMetadata.thoughtsTokenCount', null),
                 ),
                 meta: new Meta(
                     id: data_get($data, 'id', ''),
--- a/src/Providers/Gemini/Handlers/Structured.php
+++ b/src/Providers/Gemini/Handlers/Structured.php
@@ -64,6 +64,9 @@ class Structured
                         'temperature' => $request->temperature(),
                         'topP' => $request->topP(),
                         'maxOutputTokens' => $request->maxTokens(),
+						'thinkingConfig' => array_filter([
+							'thinkingBudget' => $providerOptions['thinkingBudget'] ?? null,
+						]),
                     ]),
                     'safetySettings' => $providerOptions['safetySettings'] ?? null,
                 ])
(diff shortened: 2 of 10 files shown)


ChrisB-TL:
Thank you!

Re: default, I think we should be consistent across providers if we can. For Anthropic, we have it default disabled so I'd be tempted to do the same here. (IMO its probably the more sensible default anyway?)

src/ValueObjects/Usage.php
```diff
@@ -10,6 +10,7 @@ public function __construct(
         public int $promptTokens,
         public int $completionTokens,
         public ?int $cacheWriteInputTokens = null,
-        public ?int $cacheReadInputTokens = null
+        public ?int $cacheReadInputTokens = null,
+		public ?int $thoughtTokens = null,
```
ChrisB-TL, line 14: I'd be tempted to add the thoughtTokens to the `Response` additionalContent property, as it is unique to Gemini. 

What do you think @sixlive?

tests/Providers/Gemini/GeminiTextTest.php
```diff
@@ -440,3 +441,30 @@
         expect($response->usage->cacheReadInputTokens)->toBe(88759);
     });
 });
+
+describe('Thinking Mode for Gemini', function (): void {
```
ChrisB-TL, line 445: Please can you add Http assertions that the correct payload is sent also? Conscious we don't have this enough in the code base (on the todo to add more), and we've been bitten by it a couple of times in refactors.

src/Providers/Gemini/Handlers/Text.php
```diff
@@ -81,6 +81,9 @@ protected function sendRequest(Request $request): ClientResponse
                 'temperature' => $request->temperature(),
                 'topP' => $request->topP(),
                 'maxOutputTokens' => $request->maxTokens(),
+				'thinkingConfig' => array_filter([
```
ChrisB-TL, line 84: This will be an empty array if not set. Can we instead add the ternary to the `thinkingConfig` level?

src/Providers/Gemini/Handlers/Structured.php
```diff
@@ -64,6 +64,9 @@ public function sendRequest(Request $request): array
                         'temperature' => $request->temperature(),
                         'topP' => $request->topP(),
                         'maxOutputTokens' => $request->maxTokens(),
+						'thinkingConfig' => array_filter([
```
ChrisB-TL, line 67: This will be an empty array if not set. Can we instead add the ternary to the `thinkingConfig` level?

src/Providers/Gemini/Handlers/Stream.php
```diff
@@ -242,6 +242,9 @@ protected function sendRequest(Request $request): Response
                             'temperature' => $request->temperature(),
                             'topP' => $request->topP(),
                             'maxOutputTokens' => $request->maxTokens(),
+							'thinkingConfig' => array_filter([
```
ChrisB-TL, line 245: This will be an empty array if not set. Can we instead add the ternary to the `thinkingConfig` level?

ChrisB-TL (changes requested):

src/ValueObjects/Usage.php
btumbleson: I could go both ways, but I think it's likely that all major models will converge on having "thinking" be a capability of their model, so my thought with this here was to future proof a bit.

tests/Providers/Gemini/GeminiTextTest.php
btumbleson: Good call out. Actually found a bug while adding these tests!

src/Providers/Gemini/Handlers/Text.php
```diff
@@ -81,6 +81,11 @@ protected function sendRequest(Request $request): ClientResponse
...
+					'thinkingBudget' => array_key_exists('thinkingBudget', $providerOptions)
+						? $providerOptions['thinkingBudget']
+						: null,
+				], fn($v) => $v !== null),
```
btumbleson, line 88: This is generally messier than I wanted it to be, but just a simple `array_filter` will throw away a `0` value, which is actually valid.

tests/Providers/Gemini/GeminiTextTest.php
```diff
@@ -440,3 +441,49 @@
...
+			->using(Provider::Gemini, 'gemini-2.5-flash-preview')
+			->withPrompt('Explain the concept of Occam\'s Razor and provide a simple, everyday example.')
+			->asText();
+
+		expect($response->usage->thoughtTokens)->toBe(1209);
+
+		Http::assertSent(function (Request $request) {
+			$data = $request->data();
+
+			expect($data['generationConfig'])->not->toHaveKey('thinkingConfig');
```
btumbleson, line 459: Note this is actually purposeful inline with the default convention that we would expect to see `thoughtTokens` in the response without specifying a `thinkingConfig`

src/ValueObjects/Usage.php
ChrisB-TL: I've had a look at the other providers that have reasoning models. As always with providers... its a pretty 50:50 split whether they add thinking tokens:
- OpenAI, Gemini do, and xAI do; and
- Anthropic and Deepseek don't.
- Groq and Ollama - not sure.

I reckon there's enough that do to warrant adding.

ChrisB-TL:
> Thank you!
> 
> Re: default, I think we should be consistent across providers if we can. For Anthropic, we have it default disabled so I'd be tempted to do the same here. (IMO its probably the more sensible default anyway?)

Sorry, you may have missed this comment as I didn't include in the body of my review!

btumbleson:
> > Thank you!
> > 
> > Re: default, I think we should be consistent across providers if we can. For Anthropic, we have it default disabled so I'd be tempted to do the same here. (IMO its probably the more sensible default anyway?)
> 
> Sorry, you may have missed this comment as I didn't include in the body of my review!


I can make this change, but have two hesitations:

1. This functionality is enabled via the `withProviderOptions` which somewhat implies it's provider-specific. Deviating from how Google handles this (which is on by default) feels counter-intuitive. If the method was `withThinkingBudget` then I think I'd agree with the need to better standardize against all providers. 
2. Setting the thinking budget to 0 by default will require a `ThinkingModeResolver` which as of today, only supports a single model. This will introduce incremental maintainence. Not major,  but I'm conscious of this. Similar to how earlier iterations of Prism, there was a  `StructuredModeResolver` in more providers but now has gotten removed. 

Just let me know your thoughts on how we want to proceed.

btumbleson:
@sixlive - anything you're looking to see here to merge?

sixlive:
Just need to sit down and give it a thorough review. Probably later today or tomorrow.

pushpak1300:
So @btumbleson if i don't specifically set the thinking budget the request will be sent without thinking budget ? Right?

btumbleson:
> So @btumbleson if i don't specifically set the thinking budget the request will be sent without thinking budget ? Right?

Other way around. If you don't specify a thinking budget on a model that support it (e.g. Gemini 2.5 Flash Preview), then it will include thinking. This matches Gemini's default behavior: https://ai.google.dev/gemini-api/docs/thinking#use-thinking-models

The "thoughts" themselves are not returned via the API, only a count of the tokens used. As thinking tokens are billed at a higher rate, you may not want thinking enabled, or may wish to limit the amount of tokens that can be used to "think". Both of which you can accomplish with setting `thinkingBudget`

sixlive (approved):
SICK!!! Thank you so much!

PR #347, opened 2025-04-30
refactor(ollama): enhance options handling in handlers

bradenkeith:
Whats up TJ!

Great work on this package. Finally got around to a new LLM project and decided to give it a go vs rolling my own. 

So far so good. 

I noticed Ollama options were missing, and added those with this PR.

Now users can do something like

```php
$response = Prism::text()
    ->usingTemperature(0.5)
    ->withMaxTokens(1000)
    ->withProviderOptions([
        'stop' => ['<custom_stop_sequence>'],
    ])
```

To create a custom stop sequence. Other [options are available](https://github.com/ollama/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values).

The way it's implemented, `withProviderOptions` will take priority over any of the temperature, top_p, or token configurations, should the user manually define these in the `withProviderOptions`. This would be preferential, so users could use one definition for all their options, if they so choose.

I added options to embeddings, since Ollama does make it available in their API specs.

This would close Issue #110

### DIFF
--- a/src/Providers/Ollama/Handlers/Embeddings.php
+++ b/src/Providers/Ollama/Handlers/Embeddings.php
@@ -51,6 +51,7 @@ class Embeddings
             [
                 'model' => $request->model(),
                 'input' => $request->inputs(),
+                'options' => array_filter($request->providerOptions()),
             ]
         );
     }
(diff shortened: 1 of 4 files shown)


sixlive (approved):
This is awesome, thank you so much!

sixlive:
Could you do me a HUGE favor and add a quick note to the Ollama docs about being able to use `providerOptions` as a way to configure Ollama options?

bradenkeith:
Sorry for just seeing this brother. Looks like you got it. 👍  @sixlive

PR #353, opened 2025-05-08
Fix Ollama Error when no options are provided.

axyr:
See https://github.com/prism-php/prism/issues/352


sixlive (approved):
Thank you so much!!!

PR #356, opened 2025-05-14
feat(openai): add metadata

denisdulici:
## Description

Adds key-value pair `metadata` feature for [OpenAI](https://platform.openai.com/docs/api-reference/chat/create#chat-create-metadata).

### DIFF
--- a/src/Concerns/ConfiguresModels.php
+++ b/src/Concerns/ConfiguresModels.php
@@ -12,6 +12,8 @@ trait ConfiguresModels
 
     protected int|float|null $topP = null;
 
+    protected ?array $metadata = null;
+
     public function withMaxTokens(?int $maxTokens): self
     {
         $this->maxTokens = $maxTokens;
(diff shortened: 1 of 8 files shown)


sixlive:
What would we be sending as metadata? This is kind of what the `withProviderOptions` for this I think – https://prismphp.com/providers/openai.html#provider-specific-options

denisdulici:
It is used to store additional information about the object in a structured format, and querying for objects via API or the dashboard:

image

TBH, I don't know if it's used by other models or not. I can make it available only for `OpenAI` handlers if you think so:

```php
'temperature' => $request->temperature(),
'top_p' => $request->topP(),
'metadata' => (array) $request->providerOptions('metadata'),
'tools' => ToolMap::map($request->tools()),
```

ChrisB-TL:
I haven't seen this with any other provider - so I think setting via `providerOptions` is the way to go.

sixlive:
I agree with @ChrisB-TL. Lets move this to `providerMeta` and I'll get it merged!

PR #357, opened 2025-05-14
feat: Prevent Truncated Exception Messages

### LINKED ISSUE #351: Truncated Exception Messages
Prism uses the Laravel HTTP client which by default will truncate exception messages to 120 characters. This can be annoying when you get an exception and the error message from the provider is truncated. It took me awhile to figure out how to see the entire error message which had helpful information. 

Example truncated exception message:
```
Sending to model (anthropic.claude-3-5-sonnet-20241022-v2:0) failed: HTTP request returned status code 400:
{"message":"Invocation of model ID anthropic.claude-3-5-sonnet-20241022-v2:0 with on-demand throughput isn’t supported (truncated...)
```

I figured out I can call `RequestException::truncateAt(1000)` or `RequestException::dontTruncate();` before calling Prism and then I got the full exception message which is much more helpful:
```
Sending to model (anthropic.claude-3-5-sonnet-20241022-v2:0) failed: HTTP request returned status code 400:
{"message":"Invocation of model ID anthropic.claude-3-5-sonnet-20241022-v2:0 with on-demand throughput isn’t supported. Retry your request with the ID or ARN of an inference profile that contains this model."}
```

If we think this is a concern, we could add a `truncate_request_exceptions` config option to `prism.php` where it could be set to `null` to use the default behavior, `false` to disable truncation, or an integer value to specify the number of characters.  At the very least I wanted to create this issue so if other users encounter it then they would see the solution on how to prevent the truncation.

ejunker:
Closes #351 

Prevents truncated HTTP request exception messages so you can see the entire message from the provider.

I wasn't sure of the best place to call `RequestException::dontTruncate()` so I added it to PrismServiceProvider since that is one of the few global places that didn't require adding it to each Prism Provider (Anthropic.php, OpenAI.php, etc.)

Note: when calling `RequestException::dontTruncate()` it sets it for all requests made by the Laravel HTTP client even those made outside of Prism. There does not appear to be a way to set the truncation per request.


ejunker:
I didn't realize the `RequestException::dontTruncate()` method was recently added. I added a `method_exists()` check.

sixlive:
> Note: when calling RequestException::dontTruncate() it sets it for all requests made by the Laravel HTTP client even those made outside of Prism. There does not appear to be a way to set the truncation per request.

Hmmm... I don't know if I like that... I wonder...

- Should we try and open a PR to try and do it per request?
- Should we just document this somewhere instead?
- Do we feel fine with just doing this anyway?

ChrisB-TL:
IMO, I think leave this out of Prism as it is easy enough for the developer to do in one of their app providers.

ijpatricio:
It took me a while, and some frustration to figure out about this.

I'm not so sure it wouldn't be something I would take care in the package.
Before request, disable truncation, then re-enable again would be a good sensible default.

Anyway, I'm not trying to change anyone's mind, what I think it really deserves is a note somewhere in the Docs.
I wasn't aware of this native feature, so I deep searched on Prism's docs first, dived into sour-code, only then GH issues...

@sixlive  Would you accept a PR oto the docs?

PR #368, opened 2025-05-24
feat: telemetery

sixlive:
## Description

- Adds tracing telemetry to Prism 
  - Request
  - Provider HTTP Requests
  - Tool Use
- Support for OpenTelemetry via https://github.com/prism-php/opentelemetry  

### Jaeger UI (OpenTelemetry)

image

### Log Driver

image


MrCrayon:
Would you consider using Events instead of decorator Pattern?

As example `asEmbeddings` method would become something like:

```php
// src/Embeddings/PendingRequest.php

public function asEmbeddings(): \Prism\Prism\Embeddings\Response
{
    if ($this->inputs === []) {
        throw new PrismException('Embeddings input is required');
    }

    $request = $this->toRequest();

    $traceId = Str::uuid()->toString();

    event(new RequestEvent($request, $traceId, microtime(true)));

    $response = $this->provider->embeddings($request);

    event(new ResponseEvent($response, $traceId, microtime(true)));

    return $response;
}
```

Decoupling tracing from the calls make it easier to replace tracing system and deciding what to trace, it could be even processed asyncronously.

P.S.
Thanks for the package

sixlive:
@MrCrayon I remixed telemetry using events https://github.com/prism-php/prism/pull/382 take a look and let me know what you think

sixlive:
Closing in favor of #382

PR #379, opened 2025-05-28
fix(gemini): tool property mapping

### LINKED ISSUE #350: ObjectSchema should not include additionalProperties by default or allow disabling it completely
When using Prism\Prism\Schema\ObjectSchema with Gemini, Prism sends an additionalProperties key in the JSON schema, which causes Gemini to fail with:

`Invalid JSON payload received. Unknown name "additionalProperties" at 'tools.function_declarations[0].parameters.properties[0].value'`

ChrisB-TL:
## Description

Fixes #350 and #375.


tests/Providers/Gemini/ToolTest.php
```diff
@@ -5,22 +5,46 @@
...
+        ->withObjectParameter(
+            'options',
+            'additional options',
+            [
+                new StringSchema('option1', 'description for option1'),
+            ]
+        )
         ->using(fn (): string => '[Search results]');
 
+    ray(ToolMap::map([$tool]));
```
sixlive, line 25: Can probably remove this.

sixlive (changes requested):
Looks good. Need to remove debugging

tests/Providers/Gemini/ToolTest.php
ChrisB-TL: Whoops!

sixlive (approved):

Issue #56, opened 2024-11-09
Adding debug support for request and response.


pushpak1300:
Is there a way to get Raw Json response from the ProviderResponse ?
Also It will be great to see what Request is being sent from the prism to different provider ?

sixlive:
There isn't anything right now but I'm open to it.

Do you have a specific use case in mind?

pushpak1300:
Yes. For debugging Having something like this will help a lot.

```PHP
$response = Prism::text()
    ->using(Provider::Anthropic, 'claude-3-sonnet')
    ->withSystemPrompt(view('prompts.system'))
    ->withPrompt('Explain quantum computing to a 5-year-old.')
    ->generate()
   ->dumpRawResponse(); // will  dump response in array format

   
$response = Prism::text()
    ->using(Provider::Anthropic, 'claude-3-sonnet')
    ->withSystemPrompt(view('prompts.system'))
    ->withPrompt('Explain quantum computing to a 5-year-old.')
    ->generate()
   ->ddRawResponse(); // will die& dump response in array format

$response = Prism::text()
    ->using(Provider::Anthropic, 'claude-3-sonnet')
    ->withSystemPrompt(view('prompts.system'))
    ->withPrompt('Explain quantum computing to a 5-year-old.')
   ->ddRawRequest(); //Will dump request payload and url in array format
   
  ```
What do you think ? (Similar to [this](https://laravel.com/docs/11.x/queries#debugging) )

sixlive:
I think this is totally doable. I'm working my way through a pretty major refactor in preparation for structured output / object support. I'll try and keep this in mind and we can circle back to it. Going to leave this issue open as a reference.

code-distortion:
Hi there. `dump` and `dd` variants seem useful to me too, similar to those in the Eloquent QueryBuilder, for quick verification.

I'd like to also suggest callbacks, which would allow for logging of requests/responses.

Or perhaps the middleware in the roadmap will allow for this?

pushpak1300:
I think extension of this is to have a system architected in a such way where so we can integration with system like  Langfuse  or debuggeres where we can log the request and response asynchornously. https://api.reference.langfuse.com/#get-/api/public/comments

akulmehta:
This would be really useful.

vesper8:
There still isn't any way to get a raw response that I could easily save to a database for later reviewing?

MrCrayon:
[...]

PR #382, opened 2025-05-29
feat: Telemetry

sixlive:
## Description

This PR enables telemetry for Prism requests and is a remix of https://github.com/prism-php/prism/pull/368 that uses Laravel events instead of using a decorator abstraction.

> [!NOTE]
> This also enables the [prism-php/opentelemetry](https://github.com/prism-php/opentelemetry) package.

### DIFF
--- a/src/Providers/Anthropic/Handlers/Stream.php
+++ b/src/Providers/Anthropic/Handlers/Stream.php
@@ -11,6 +11,7 @@ use Illuminate\Http\Client\Response;
 use Illuminate\Support\Arr;
 use InvalidArgumentException;
 use Prism\Prism\Concerns\CallsTools;
+use Prism\Prism\Concerns\TracksHttpRequests;
 use Prism\Prism\Enums\ChunkType;
 use Prism\Prism\Exceptions\PrismChunkDecodeException;
 use Prism\Prism\Exceptions\PrismException;
@@ -31,7 +32,7 @@ use Throwable;
 
 class Stream
 {
-    use CallsTools, HandlesResponse;
+    use CallsTools, HandlesResponse, TracksHttpRequests;
 
     protected StreamState $state;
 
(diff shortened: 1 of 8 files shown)


MrCrayon:
There is another thing in the tracing to note: `text_generation` is enclosing all process even if technically the first call to the  LLM ends before calling the Tool and there is another call to LLM with Tool result.

- text_generation
  - http_request
  - http_request
  - tool_call
    - embeddings
    - embeddings
  - tool_call
- text_generation

I think this could be ok if we track all `http_requests`, right now it looks like `embeddings` call and `text_generation` call done after tool call are not tracked.

MrCrayon:
@sixlive I am taking some time to understand the code base properly.
As a first step I think we could simplify Http calls tracing having a Trait that initializes Http client.
I created #386 for that purpose.

Using that we could then add tracing in the Trait like this:

```php
return Http::when($token !== '', fn ($client) => $client->withToken($token))
    ->when($headers !== [], fn ($client) => $client->withHeaders($headers))
    ->withRequestMiddleware(
        function (RequestInterface $request) {
            Event::dispatch(new HttpRequestStarted(
                contextId: $this->contextId,
                method: $request->getMethod(),
                url: (string) $request->getUri(),
                provider: $this->provider->value,
                parentContextId: $this->parentContextId,
                attributes: json_decode((string) $request->getBody(), true),
            ));

            return $request;
        }
    )
    ->withResponseMiddleware(
        function (ResponseInterface $response) {
            Event::dispatch(new HttpRequestCompleted(
                contextId: $this->contextId,
                statusCode: $response->getStatusCode(),
                attributes: json_decode((string) $response->getBody(), true),
            ));

            return $response;
        }
    )
    ->withOptions($options)
    ->when($retry !== [], fn ($client) => $client->retry(...$retry))
    ->baseUrl($baseUrl);
```

sending the raw request and response makes it available for logging closing #56.
You can choose what to keep in the Listener included with the package.

sixlive:
So I like the idea of events but I think linking spans and parent-child relationships are going to become difficult so I'm going to roll with a different pattern via https://github.com/prism-php/prism/pull/389

MrCrayon:
@sixlive can you have a look at this [implementation](https://github.com/MrCrayon/prism/commits/test-events) check the last commit [Add tracing](https://github.com/MrCrayon/prism/commit/3153a939967b5bd260e22e64ac1f1df454946fdc)

I am using Laravel `Context` to keep track of the operations in the stack.

It's not connected to the logging driver, it's just the base to use events.

This is a RAG flow with the Tool that calls embeddings and reranks, the first `Trace::begin` it's optional to give a `parentId`.
```php
Trace::begin('run #1', () => optionalCallEventOrDoSomething());

Prism::text()
    ->using(Provider::Anthropic, model: 'claude-sonnet-4-20250514')
    ->withMessages($conversation->toArray())
    ->withMaxSteps(3)
    ->withTools([new SearchTool()])
    ->asText();

Trace::end();
 ```
 
 With a Listener that logs all calls it produces this track:
```
text started
    http started
    http completed
text completed
tool_call started
    embeddings started
        http started
        http completed
    embeddings completed
    reranks started
        http started
        http completed
    reranks completed
tool_call completed
text started
    http started
    http completed
text completed
```

Right now when a tool is used is tracking properly only with `Anthropic` `text`, the other text handlers need to be updated.

PR #398, opened 2025-06-07
Fix: openai tool call map fix #397

### LINKED ISSUE #397: OpenAI Error when `choices.0.message.tool_calls` is null instead of an array
I'm using an OpenAI compatible api via LiteLLM. Trying to use a tool and running into an issue where `Prism\Prism\Providers\OpenAI\Maps\ToolCallMap` is expecting an array as an argument. 

This is causing an error because `choices.0.message.tool_calls` is actually null in the response.

`data_get` only seems to return the default if the key doesn't exist. 

A real easy fix would be something like this. 

```php
class ToolCallMap
{
    /**
     * @param  ?array<int, array<string, mixed>>  $toolCalls
     * @return array<int, ToolCall>
     */
    public static function map(?array $toolCalls): array
    {
        if ($toolCalls === null) {
            return [];
        }

        return array_map(fn (array $toolCall): ToolCall => new ToolCall(
            id: data_get($toolCall, 'id'),
            name: data_get($toolCall, 'function.name'),
            arguments: data_get($toolCall, 'function.arguments'),
        ), $toolCalls);
    }
}
```

This is how I'm using it.

```php
        $tool = Tool::as('determineGradeScore')
            ->for('This function will return the readability of the edited text, after the GPT has made changes.')
            ->withStringParameter('editedText', 'The current edited text, after the GPT has made changes.')
            ->withStringParameter('formula', "The formula used to determine the readability of the text. Must be one of the following: $formulas.")
            ->using(fn (string $editedText, string $formula) => (string) $textStatistics->{$formula}($editedText));

        $response = Prism::text()
            ->using(Provider::OpenAI, 'gpt-4.1-mini')
            ->withSystemPrompt(view('prompts.readability', ['format' => ReadabilityFormat::from($format)]))
            ->withPrompt("Edit the following text to be more readable for a {$grade} grade student: \n\n".$content)
            ->withMaxSteps(2)
            ->withTools([$tool])
            ->withToolChoice(ToolChoice::Auto)
            ->asText();

```

myleshyson:
## Description
This address the issue in #397. Sometimes, depending on the OpenAI api, `choices.0.message.tool_calls` is actually null instead of an array.

### DIFF
--- a/src/Providers/OpenAI/Maps/ToolCallMap.php
+++ b/src/Providers/OpenAI/Maps/ToolCallMap.php
@@ -9,11 +9,15 @@ use Prism\Prism\ValueObjects\ToolCall;
 class ToolCallMap
 {
     /**
-     * @param  array<int, array<string, mixed>>  $toolCalls
+     * @param  ?array<int, array<string, mixed>>  $toolCalls
      * @return array<int, ToolCall>
      */
-    public static function map(array $toolCalls): array
+    public static function map(?array $toolCalls): array
     {
+        if ($toolCalls === null) {
+            return [];
+        }
+
         return array_map(fn (array $toolCall): ToolCall => new ToolCall(
             id: data_get($toolCall, 'id'),
             name: data_get($toolCall, 'function.name'),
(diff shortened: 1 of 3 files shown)


sixlive (approved):
Thanks!

PR #406, opened 2025-06-12
fix: change input_text to text for user message type

kauffinger:
This fixes streaming responses for openai.

I am not sure if I am crazy. Without this, I get the following error when streaming responses:

```
  "error" => array:4 [▼
    "message" => "Invalid value: 'input_text'. Supported values are: 'text', 'image_url', 'input_audio', 'refusal', 'audio', and 'file'."
    "type" => "invalid_request_error"
    "param" => "messages[1].content[0].type"
    "code" => "invalid_value"
  ]
```

But the docs [state otherwise](https://platform.openai.com/docs/api-reference/responses/list). Normal responses work with `input_text`. With `text`, both works.

PR #408, opened 2025-06-12
feat: stream openai via responses api

kauffinger:
Started implementing streaming via Responses API for OpenAI, as #406 is just a temporary fix (or might brake something else).

Had an agent running to do most of the work, I'll clean up when I get to it.

2 tests failing, still.


kauffinger:
- multiple tool calls with reasoning models still fail

MrCrayon:
@kauffinger I would move `handleToolCalls` after the stream is fully read otherwise you keep the stream open while doing tool calls.

The logic would be:
```php
protected function processStream(Response $response, Request $request, int $depth = 0): Generator
{
    // remove depth check and exception, moved inside handleToolCalls

    while (! $response->getBody()->eof()) {
        //process stream
    }

    //if there are tool calls call handleToolCalls
}

protected function handleToolCalls(Request $request, string $text, array $toolCalls, int $depth): Generator
{
    // Make tool calls

    $depth++;

    // check of depth is done here because here the new request starts
    if ($depth < $request->maxSteps()) {
        $nextResponse = $this->sendRequest($request);
        yield from $this->processStream($nextResponse, $request, $depth);
    }
}

File: src/Providers/OpenAI/Handlers/Stream.php [hunk 1]
```diff
@@ -50,37 +57,50 @@ public function handle(Request $request): Generator
...
                 $toolCalls = $this->extractToolCalls($data, $toolCalls);
 
                 continue;
             }
 
-            // Handle tool call completion
-            if ($this->mapFinishReason($data) === FinishReason::ToolCalls) {
-                yield from $this->handleToolCalls($request, $text, $toolCalls, $depth);
+            if ($this->isFinalEvent($data)) {
+                if ($toolCalls !== []) {
```
MrCrayon, line 84: This should be removed as well since `handleToolCalls` is moved to the end after `while` is done.
I don't think there should be any early return, yield chunks when needed and let stream finish.

src/Providers/OpenAI/Handlers/Stream.php
```diff
@@ -123,17 +147,49 @@ protected function parseNextDataLine(StreamInterface $stream): ?array
...
+            foreach ($toolCalls as &$call) {
+                if (($call['id'] ?? null) === $callId) {
+                    $call['arguments'] .= $delta;
+                    break;
+                }
             }
 
-            $arguments = data_get($toolCall, 'function.arguments');
+            return $toolCalls;
+        }
```
pushpak1300, line 180: Do you think this is necessary ?

as argument will be handled when `type='response.function_call_arguments.done'`

kauffinger:
Ok, so I've now thrown out all the hallucinated and useless stuff. There is still room for improvement, but we're at least not in agent slop land anymore.

Everything works fine for now, except for o4-mini. We currently do not pass any reasoning items back to OAI and it seems we need to.

This leads me to a new issue. I've dabbled in this and tried to just use the current tool call pathway used in the TextHandler. That doesn't work, as the tool obviously does not exist.

I think we need a way to handle reasoning content and pass it as context.
You can see the required Datastructure the following way (this doc page drives me nuts): https://platform.openai.com/docs/api-reference/responses/create

Then, navigate (by clicking `show possible types` to: `input` -> `input item list` -> `item` ->  `reasoning`.

I think we might need a new Value Object `Reasoning` which is then passed on the following calls. It's as far as I can tell not part of an AssistantMessage. Also, we can then send the reasoning summary chunks through the stream.

I'll be blocked for a bit again, so I won't be able to work on this immediately.

@sixlive / anyone with experience with passing back reasoning to OAI, what do you think?

sixlive:
This seems relevant https://platform.openai.com/docs/guides/reasoning?api-mode=responses#keeping-reasoning-items-in-context

pushpak1300 (commented):
@sixlive  this looks good. 

I'll test it locally and give heads up.

kauffinger:
Looks good to me, too - will only be able to test later today.

One thing I haven't tried yet is o3-pro. That's the first model that doesn't support chat completions at all. Worth making sure this works.

kauffinger:
Well, I can't test it, as I first need to go through the process of verifying an organization of mine. But If I had to guess, I'd say it works.

zangopushpak (approved):

pushpak1300 (approved):

sixlive (approved):

PR #409, opened 2025-06-12
fix(openai): streaming output response map

sixlive:
## Description

This PR fixes a bug introduced with #256


cade-may:
getting the following issue when using `->asStream()`

```
[2025-06-13 00:17:28] local.ERROR: Detailed API Error Response {"status":404,"response_body":"{
  \"error\": {
    \"message\": \"This is not a chat model and thus not supported in the v1/chat/completions endpoint. Did you mean to use v1/completions?\",
    \"type\": \"invalid_request_error\",
    \"param\": \"model\",
    \"code\": null
  }
```

I assume this is being resolved via https://github.com/prism-php/prism/pull/408, but just wanted to provide some additional context :) 

thank you guys for the awesome framework

kauffinger:
Ah yeah, that's the easier way which doesn't touch anything. Good stuff.


@cade-may yeah, you will need to wait if it works on asText but not with asStream :) I'll try to get it done until next week, but no promises.

sixlive:
@cade-may what model where you using?

cade-may:
> @cade-may what model where you using?

o3-pro-2025-06-10 

that might be a bit of an edge case but it would be awesome if we could support o3 pro if possible

thank you guys :)

PR #415, opened 2025-06-16
fix: ensure tool calls still work with completions api

kauffinger:
Tried 70.1 today in prod and realized tool calls don't work. Quickly rolled back all changes from #256 by using the same pattern as with the `ChatMessageMap` in #409 

Depending on how quick we are with #408 this might be worth merging in between.


sixlive:
I'll merge #408 today

kauffinger:
perfect

PR #421, opened 2025-06-18
feat(openai): Response meta while streaming

danharrin:
## Description

Uses the existing `Meta` object and chunk type to expose information about the response ID and model from the API response. Otherwise, it's impossible to infer the `response_id` from the response to use in the next completion (as `previous_response_id`).


sixlive (approved):
Sick! Thanks!

PR #434, opened 2025-06-23
refactor: make provider contract abstract

ChrisB-TL:
## Description

v0.75.0 and other versions that add new methods to the provider contract are breaking changes for custom providers (including Bedrock and other "official" providers that are later built as separate packages).

This PR makes the provider contract an abstract class, with sensible defaults for all current methods. 

Aside from reducing breaking changes, it also de-clutters the providers a little.

This could equally just be a regular class, though (1) I think abstract sends the right "signal"; and (2) it future proofs for adding methods that cannot have sensible defaults in the future.

P.S. I admit the irony that this in itself will be a breaking change for custom providers - but hopefully one of the last!

## Breaking Changes
- Provider relocated from Prism\Prism\Contracts to Prism\Prism\Providers
- Custom providers need to extend Provider instead of implementing it


src/Contracts/Provider.php
```diff
@@ -16,20 +18,46 @@
...
+    }
+
+    public function handleRequestExceptions(string $model, Throwable $e): never
+    {
+        if ($e instanceof PrismException) {
+            throw $e;
+        }
+
+        if (! $e instanceof RequestException) {
+            throw PrismException::providerRequestError($model, $e);
```
ChrisB-TL, line 58: I've done this as per the provider methods that have been added.

However, I think it might make more sense to have the default just throw the exception "as is" - if it is not a RequestException?

src/Contracts/Provider.php
sixlive, line 1: We probably need to move this out of the contract namespace

sixlive (approved):
Sick, thanks for this, nice refactor!

MrCrayon:
Thanks @ChrisB-TL you did a great job to improve the refactor, much better now.

I have a question: in `Provider` abstract class it's defined `handleRequestException` method but then every provider that does not implement a custom `handleRequestException` still uses the `Trait`.

```php
class Ollama extends Provider
{
    use HandlesRequestExceptions;
```
Do you think it would be better to:

 1. Remove the Trait and enhance the default implementation in `Provider` abstract with:

```php
    public function handleRequestException(string $model, RequestException $e): never
    {
        match ($e->response->getStatusCode()) {
            413 => throw PrismRequestTooLargeException::make(self::class),
            429 => throw PrismRateLimitedException::make([]),
            529 => throw PrismProviderOverloadedException::make(self::class),
            default => throw PrismException::providerRequestError($model, $e),
        };
    }
```

 2. Fix the Trait that is still using `Throwable` as parameter and doing a couple of checks that are not needed if we only accept `RequestException`.

3. Other solution?

I would go with the first one.

ChrisB-TL:
Good catch - I missed the trait. I'll do the first and PR it now.

PR #437, opened 2025-06-23
feat(openai): yield thinking chunks

kauffinger:
## Description

Finally got around to verify my org. This adds streaming for thinking chunks.

## Breaking Changes
Might be breaking in the sense that if people don't check for the chunk type, they will now render thinking chunks.

### DIFF
--- a/src/Providers/OpenAI/Handlers/Stream.php
+++ b/src/Providers/OpenAI/Handlers/Stream.php
@@ -74,6 +74,20 @@ class Stream
                 continue;
             }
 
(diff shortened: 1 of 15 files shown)


sixlive:
I'd prefer to be explicit about what keys are used, at least for now

kauffinger:
Sure thing! Did it because I (stupidly) didn't realize this was the case and thought my org still wasn't verified 🙃

sixlive (approved):

PR #443, opened 2025-06-24
chore: add example to custom provider and apply style guide

kauffinger:
## Description
I took a look at the docs and saw the old version of the custom provider docs, before #434
So, I ran claude code with the style guide and instructions to align it with the new PR, only to find out it was already aligned, just not deployed.

I still like the style additions, examples, and the tips claude added so much that I decided to open this PR anyways.

## Breaking Changes
none, docs

### DIFF
--- a/docs/advanced/custom-providers.md
+++ b/docs/advanced/custom-providers.md
@@ -2,19 +2,45 @@
 
 Want to add support for a new AI provider in Prism? This guide will walk you through creating and registering your own custom provider implementation.
 
-## Provider Interface
+## Building Your Provider
 
-All providers must extend the `Prism\Prism\Providers\Provider` class.
+All providers must extend the `Prism\Prism\Providers\Provider` abstract class. This base class provides default implementations for all required methods, throwing exceptions for unsupported actions.
 
-The abstract class has a default method for all current required methods, though this may change.
+When creating your provider, you'll only need to override the methods for the features you want to support:
+- `text()` - For text generation
+- `structured()` - For structured output generation
+- `embeddings()` - For creating embeddings
+- `images()` - For image generation
+- `stream()` - For streaming text responses
 
-Each provider should:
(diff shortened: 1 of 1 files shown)


docs/advanced/custom-providers.md
```diff
@@ -2,19 +2,45 @@
...
+When creating your provider, you'll only need to override the methods for the features you want to support:
+- `text()` - For text generation
+- `structured()` - For structured output generation
+- `embeddings()` - For creating embeddings
+- `images()` - For image generation
+- `stream()` - For streaming text responses
 
-Each provider should:
-- Overwrite the methods for the actions it supports. 
-- Overwrite the `handleRequestExceptions` method if supports advanced request exceptions.
```
MrCrayon, line 13: I would keep this line refactored since it's one of the methods that can be overridden if needed.
It was moved in Best Practices section but in a more generic way.

kauffinger: Hmm, I'm not sure if I'd include it at the top where the 'feature methods' are. I think creating a separate Error Handling section / making the line in Best Practices more explicit is better for what the reader wants to accomplish when on the page.

docs/advanced/custom-providers.md
```diff
@@ -58,3 +84,43 @@ $response = Prism::text()
...
+                    ? null
+                    : (int) $e->response->header('retry-after'),
+            ),
+            default => parent::handleRequestException($model, $e),
+        };
+    }
+}
+```
+
+The method must throw an exception (return type `never`). If you don't handle a specific status code, make sure to call the parent method to maintain the default error handling.
```
kauffinger, line 116: What do you think of this @MrCrayon

MrCrayon: Looks good to me, nice suggestion about calling the parent method by default when extending!

sixlive (approved):

PR #447, opened 2025-06-25
feat(gemini): add video/audio/Youtube link media support

uchm4n:
## Description
Added media support for videos, audio, and YouTube links for **Gemini**
- Added `Media` value object with support for local files, storage paths, URLs, base64, and raw content
- New test fixtures: sample video/audio files and API response mocks
- Complete documentation at `/docs/input-modalities/media.md`

## Breaking Changes
NO breaking changes


ChrisB-TL:
Thank you! 

My only thought is we should probably have a separate value object for each of video and audio  extending Media.

(1) as it's clearer to the dev; and (2) because other providers may treat them differently and need separate mappers in the future.

For Gemini though could just have a AudioVideoMapper that handles both.

avengerweb:
Gemini has 20MB limit for inline files:
https://ai.google.dev/gemini-api/docs/image-understanding#image-input

If it is more than 20MB they need to be uploaded, which is also tricky.

that is junk code I used to upload files if anyone interested, it works for both images and videos. But you need to modify prism code for ImageMapper as it allows only inline right now
https://github.com/prism-php/prism/blob/main/src/Providers/Gemini/Maps/ImageMapper.php#L19
```php
    public function toPayload(): array
    {
        if ($this->media->isUrl()) {
            return [
                'file_data' => [
                    'file_uri' => $this->media->url()
                ],
            ];
        }

        return [
            'inline_data' => [
                'mime_type' => $this->media->mimeType(),
                'data' => $this->media->base64(),
            ],
        ];
    }
```
```php
    public function upload(UploadedFile|string $file, $disk = 'cloud-private'): array
    {
        if (is_string($file)) {
            $storage = Storage::disk($disk);
            $content = $storage->get($file);
            $mimeType = $storage->mimeType($file);
            $fileSize = $storage->size($file);
            $ext = pathinfo($file, PATHINFO_EXTENSION);
        } else {
            $fileSize = $file->getSize();
            $mimeType = $file->getMimeType();
            $content = $file->get();
            $ext = $file->getClientOriginalExtension();
        }

        $req = \Http::withHeaders([
            'X-Goog-Upload-Protocol' => 'resumable',
            'X-Goog-Upload-Command' => 'start',
            'X-Goog-Upload-Header-Content-Length' => $fileSize,
            'X-Goog-Upload-Header-Content-Type' => $mimeType,
            'x-goog-api-key' => config('gemini.api_key'),
            'Content-Type' => 'application/json',
        ])
            ->post('https://generativelanguage.googleapis.com/upload/v1beta/files', [
                'file' => [
                    'display_name' => 'image.'.$ext,
                ],
            ]);

        $uploadUrl = $req->header('X-Goog-Upload-Url');
        $res = \Http::withHeaders([
            'Content-Length' => $fileSize,
            'X-Goog-Upload-Offset' => 0,
            'X-Goog-Upload-Command' => 'upload, finalize',
            'x-goog-api-key' => config('gemini.api_key'),
        ])
            ->withBody($content, $mimeType)
            ->post($uploadUrl);

        $res = $res->json();

        // video will be also "processed", so we need to wait till it changes status
        if (Str::startsWith($mimeType, 'video') && $res['file']['state'] === 'PROCESSING') {
            $tries = 5;
            while ($res['file']['state'] === 'PROCESSING') {
                sleep((6 - $tries));
                $tries--;
                $res = [
                    'file' => \Http::withHeaders([
                        'x-goog-api-key' => config('gemini.api_key'),
                    ])->get($res['file']['uri'])->json(),
                ];
            }
        }

        return $res;
    }
```
```php
$url = Arr::get($uploadResp, 'file.uri', null)
$userMessage->additionalContent[] = Image::fromUrl($url);
```php

sixlive (approved):
SICK!!! Thank you so much!

PR #454, opened 2025-06-28
refactor(streaming): Dedicated chunks

sixlive:
## Description

Creates dedicated chunks for each chunk type. This will provide stronger typing and consistency for streaming outputs. 

## Breaking Changes

- Big breaking change for chunks and streaming (but I think its worth it)

### DIFF
--- a/tests/Providers/Anthropic/StreamTest.php
+++ b/tests/Providers/Anthropic/StreamTest.php
@@ -460,7 +460,7 @@ describe('thinking', function (): void {
             $chunks[] = $chunk;
         }
 
-        $thinkingChunks = (new Collection($chunks))->filter(fn ($chunk): false => $chunk instanceof ThinkingChunk);
+        $thinkingChunks = (new Collection($chunks))->filter(fn ($chunk): bool => $chunk instanceof ThinkingChunk);
 
         expect($thinkingChunks->count())->toBeGreaterThan(0);
 


avengerweb:
It feels like a step backward in terms of end-user usability and developer experience. While it’s definitely cleaner to have something like a `MetaChunk` or `TextChunk`, the issue is that—**from the AI provider’s perspective**—it’s more efficient (latency, payload size, etc.) to only send what has changed.

But as a developer building end-user applications, I’d much rather focus on business logic than spend time writing boilerplate just to handle provider-specific quirks. Otherwise, streaming starts to feel more like dealing with low-level typed DTOs than a practical abstraction layer.

Moreover, this approach **contradicts the logic of some providers**—for example, **Gemini** includes usage and metadata in *every* chunk. That means you'd need to `yield` both usage and text simultaneously to preserve accuracy, which defeats the purpose of splitting chunks into isolated types like `MetaChunk` and `TextChunk`.

So, I’d recommend thinking from the **end-library user’s** perspective. I know you're also building an AI proxy endpoint where the behavior you're describing makes sense. But here, as someone building a chat or agent experience, I’m looking for:

1. **Clear signal of turn start** (when stream opens).
2. **Clear signal of turn end** (when stream closes and no more deltas/meta/usage is expected), ideally with the full response assembled.
3. **Delta chunks** that always include a **turn/message ID**, to make production debugging across multiple parallel sessions practical.
4. **Separation of thinking/internal reasoning deltas** to distinguish structured tool outputs or logs from final messages.

With this structure, I could build persistent, debuggable streaming applications with minimal overhead—without the boilerplate.

Here’s a simple, ideal example using Laravel:

```php
foreach ($stream as $chunk) {
    if ($chunk instanceof TurnStartedChunk) {
        broadcast(new MessageStartedEvent(turnId: $chunk->id, delta: $chunk->delta));
    }

    if ($chunk instanceof DeltaTextChunk) {
        broadcast(new NewMessageDelta(turnId: $chunk->id, delta: $chunk->delta));
    }

    if ($chunk instanceof TurnFinishedChunk) {
        $assistantMessage = new Message(
            role: "assistant",
            content: $chunk->content,
            meta: $chunk->meta,
            usage: $chunk->usage
        );
        $assistantMessage->save();

        broadcast(new MessageCompletedEvent(turnId: $chunk->id, message: $assistantMessage));
    }
}
```

Otherwise, you end up writing code like this, which tries to **patch together streaming behavior manually** from partial chunks:

```php
$messages = collect();
$message = new UserChatMessages();
$message->userChat()->associate($chat);
$message->role = 'assistant';
$message->message = '';

foreach ($req->asStream() as $chunk) {
    $this->heartbeat();

    if ($chunk->chunkType === ChunkType::Text) {
        $message->message .= $chunk->text;

        if ($chunk->text !== '') {
            broadcast(new NewTextMessageChunkEvent(
                $message,
                $chunk->text,
                responseId: $message->metadata['id'] ?? null
            ));
        }
    }

    if ($chunk->chunkType === ChunkType::Meta && $chunk->meta) {
        $message->metadata = [
            ...($message->metadata ?? []),
            'id' => $chunk->meta->id,
            'model' => $chunk->meta->model
        ];
    }

    if ($chunk->usage) {
        $message->prompt_tokens = $chunk->usage->promptTokens;
        $message->completion_tokens = $chunk->usage->completionTokens;
        $message->save();
        NewChatMessageEvent::dispatch($message);
    }

    if ($chunk->chunkType === ChunkType::ToolCall) {
        $message->function_call = $chunk->toolCalls;
        $message->message = '';
        $message->save();
        NewChatMessageEvent::dispatch($message);
        $messages->push($message);
        $message = null;
    }

    if ($chunk->chunkType === ChunkType::ToolResult) {
        $message = new UserChatMessages();
        $message->userChat()->associate($chat);
        $message->role = 'tool';
        $message->message = '';
        $message->metadata = [
            'tool_results' => $chunk->toolResults
        ];
        $message->save();
        NewChatMessageEvent::dispatch($message);
        $messages->push($message);

        $message = new UserChatMessages();
        $message->userChat()->associate($chat);
        $message->role = 'assistant';
    }

    if ($chunk->finishReason && !in_array($chunk->finishReason, [FinishReason::Unknown, FinishReason::ToolCalls])) {
        if ($chunk->finishReason === FinishReason::Stop) {
            $message->save();
            $messages->push($message);
        } else {
            Log::error('AI: Bad finish reason: ' . $chunk->finishReason->name, [
                'chunk' => $chunk,
                'message' => $message,
            ]);
        }
    }
}
```

That’s a lot of glue just to save and stream assistant responses. A well-structured stream abstraction would eliminate most of it.

pushpak1300:
This changes make sense. but for sending this to UI. I really think we need to implement something like https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#to-data-stream so that we can send them efficiently to UI. In order to implement this we need this current chunk implementation for sure.  @avengerweb

avengerweb:
> This changes make sense. but for sending this to UI. I really think we need to implement something like https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#to-data-stream so that we can send them efficiently to UI. In order to implement this we need this current chunk implementation for sure. @avengerweb

Not sure how this contradicts my concerns.

I have two Laravel projects running customized versions of Prism, powering UI chat interfaces (not just APIs), each serving over 1,000 daily users. I was forced to modify Prism because we allow users to switch between different AI providers (OpenAI, Anthropic, Gemini), and they all behave differently in terms of streaming chunks and formats.

When I use a library like Prism, I expect it to bridge those differences and present a unified format, rather than exposing the provider-specific logic to my application.

In non-production/demo scenarios, it’s fine to yield all raw chunks directly to the UI. But in real-world production use cases, you typically need to:

1. Persist conversation history.
2. Handle disconnects, reconnects, and multiple parallel chat sessions (even from the same user).
3. Support retries and proper error handling.
4. Apply content filters and moderation layers.

You can still send small chunks to the frontend incrementally, but at the end you must confirm that all chunks have been processed by the backend (e.g., stored in the database). You also likely want to include timestamps, enrich the data, and make sure everything is complete before marking the message as “done.”

What I’m trying to highlight is: not all chunks need to go to the frontend, and not all chunks are relevant to the backend.
For example, the backend doesn’t care about partial deltas for storage—it only needs to know when a complete message is ready.

I don’t mind having separate MetaChunk or UsageChunk types—but I really need each TextChunk to include an ID, so I can associate it with the correct message or request in both frontend and backend logic.

sixlive:
@avengerweb https://github.com/prism-php/prism/discussions/569

PR #459, opened 2025-07-01
docs(ExceptionSupport): added designation for partial support

designgears:
## Description
added designation for partial support and changed OpenAI rate limit to partially supported.

### DIFF
--- a/docs/components/ExceptionSupport.vue
+++ b/docs/components/ExceptionSupport.vue
@@ -30,6 +30,21 @@
                 />
               </svg>
               <svg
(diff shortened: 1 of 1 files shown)


ChrisB-TL:
If not throwing a PrismRateLimitedException, this is bug rather partial support.

designgears:
> If not throwing a PrismRateLimitedException, this is bug rather partial support.

Seems to be the case, digging into the bug now. I’ll leave this PR, partial support sends like a good thing to have a designation for either way.

PR #475, opened 2025-07-07
feat: add provider tools to oai stream handler

kauffinger:
## Description
Moved the buildTools logic into a new BuildsTools trait to avoid code duplication between the Stream and Text handlers. Updated both handlers to use the trait, and added a new test and fixture to verify provider tool support in streaming responses.

## Breaking Changes
None


sixlive (approved):
Sick, thanks!!

Issue #479, opened 2025-07-08
[OpenAi] The `strict` flag when using Structured Mode can only currently be `true`


EpicnessTwo:
When using OpenAi with a Structured Schema Output, the `strict` flag should be set to `true` or `false`.

Currently inside this package, the only accepted value is `true` since setting it to `false` will change it to `null`, which is then omitted using `Arr:whereNotNull()`. 

This can be seen here: https://github.com/prism-php/prism/blob/main/src/Providers/OpenAI/Handlers/Structured.php#L122-L139

Ideally, this needs to allow both `true` and `false` instead of changing all non-true values to null.

ChrisB-TL:
Does it default to false if not set?

EpicnessTwo:
> Does it default to false if not set?

After some testing, it seems to be `true` by default when using their API

PR #481, opened 2025-07-09
bugfix: Updated logic for Strict flag in OpenAi Structured Output

EpicnessTwo:
## Description

This is a fix for #479 allowing both true and false to be valid options for `schema.strict`


pushpak1300 (approved):

sixlive:
I just merged a fix before realizing this PR existed. Thanks for the contribution!

PR #485, opened 2025-07-10
Fix tool result cache accumulation across multiple tool calls

elliottlawson:
## Summary
- Fixes issue where tool_result_cache_type was being applied to each tool result message instead of only the last one
- Moves caching logic from message creation to final mapping stage
- Ensures only the last tool result across all messages gets cached

## Changes
- Remove cache application in `Text::handleToolCalls()` and `Stream::addMessagesToRequest()` 
- Add logic to `MessageMap::map()` to identify and cache only the last tool result message
- Add comprehensive test coverage for the caching behavior

## Test plan
- [x] Added new test file `ToolResultCachingTest.php` with 3 test cases
- [x] Verified all existing Anthropic tests pass (103 tests)
- [x] Ran PHPStan static analysis - no errors
- [x] Applied code formatting with Pint and composer format


sixlive:
@ChrisB-TL can you take a look at this?

ChrisB-TL:
> Will this not result in missing the last cache on subsequent calls, as the breakpoint is no longer there? I.e. you'll miss the cache until the last call, which will undermine the point of it?
> 
> Anthropic does have a limit of four cache break points, so the current implementation will break eventually for chains with multiple tool calls.

Actually, possibly not. From the docs:

```
If any of these previous positions match cached content from earlier requests, the system uses the longest matching prefix
This means you don’t need multiple breakpoints just to enable caching - one at the end is sufficient
```

Would you be able to test out with real requests to confirm, and a test that uses recorded fixtures?

PR #487, opened 2025-07-12
feat(file)!: User can provide Document and Image objects instantiated with provider file ID

roymckenzie:
## Description
- Add support for creating Document and Image objects from Provider file IDs using static `Document::fromFileId()` and `Image::fromFileId()` methods
- Add media validation and payload creation for file IDs in DocumentMapper and ImageMapper classes
- Add support for file ID handling in Media
- Updated tests to test mapping and Document/Image fromFileId creation

## Breaking Changes
- Remove `OpenAIFile` class and `Message` mapping logic for `OpenAIFile` objects in favor of `Media::formFileId()`.


ChrisB-TL:
We implemented OpenAI's equivalent separately from Document and Image as at the time they were the only provider doing this.

It's still relatively few providers, but I suspect will grow so probably worth doing.

So, if we support file ids on Document and Image we should probably refactor OpenAI to use it as part of this PR also.

sixlive:
I'm open to this! Can you resolve the current conflicts?

roymckenzie:
@sixlive I resolved the formatting issues with Pint, however regarding the phpstan errors, I wanted to check with you.

```shell
 ------ ------------------------------------------------------------------------------------------------- 
  Line   Providers/OpenAI/Maps/MessageMap.php                                                             
 ------ ------------------------------------------------------------------------------------------------- 
  :88    Call to deprecated method files() of class Prism\Prism\ValueObjects\Messages\UserMessage:        
         Use `documents()` instead.                                                                       
         🪪  method.deprecated                                                                            
  :88    Call to deprecated method mapFileParts() of class Prism\Prism\Providers\OpenAI\Maps\MessageMap:  
         Use `mapDocumentParts()` instead.                                                                
         🪪  staticMethod.deprecated                                                                      
 ------ ------------------------------------------------------------------------------------------------- 
```

I wanted to ensure backwards compatibility for anyone using `OpenAIFile` class so I marked the `mapFileParts` method in `Providers\OpenAI\Maps\MessageMap.php` as deprecated and the `files` method in `ValueObjects\Messages\UserMessage.php` as deprecated rather than removing them and the OpenAIFile class.

However, both of these methods are called in the OpenAI `MessageMap` class.

```php
    protected function mapUserMessage(UserMessage $message): void
    {
        $this->mappedMessages[] = [
            'role' => 'user',
            'content' => [
                ['type' => 'input_text', 'text' => $message->text()],
                ...self::mapImageParts($message->images()),
                ...self::mapDocumentParts($message->documents()),
                // Now deprecated methods being called that ensure backwards compatibility
                ...self::mapFileParts($message->files()),
            ],
            ...$message->additionalAttributes,
        ];
    }
    ```
    
Should I mark this line as ignore for phpstan or should I make a breaking change that takes out all the logic for the use of `OpenAIFile` class?

Thank you for your feedback!

sixlive:
I'm okay breaking things and just removing the OpenAI file class in favor of support for a broader abstraction

PR #502, opened 2025-07-16
fix(openai): Add ability to set OpenAI provider options for embedding

jrking4:
## Description
You currently cannot set a custom dimension for an OpenAI embedding.  This pull request adds the ability to set any provider option for the OpenAI Embeddings API 
[https://platform.openai.com/docs/api-reference/embeddings](url)


pushpak1300:
this is great can you please add a small test for this ?

pushpak1300 (approved):

sixlive (approved):

PR #511, opened 2025-07-23
fix(openai): Fix and improve OpenAI audio handling

aimeos:
Uploading audio files to OpenAI doesn't work at all because of incorrect usage of the Laravel HTTP facade for multipart requests. This PR fixes the upload problem and improves handling different response format from whisper-1 API depending on the sent provider specific parameters.


sixlive:
@pushpak1300 can you take a look at this?

aimeos:
I use this code, which is almost the same but it doesn't work with Prism 0.80:
```
        $prism = Prism::audio()->using( config( 'cms.ai.audio', 'openai' ), config( 'cms.ai.audio-model', 'whisper-1' ) );
        $file = Audio::fromBase64( base64_encode( $upload->getContent() ), $upload->getMimeType() );

        $response = $prism->withInput( $file )
            ->withProviderOptions([
                'response_format' => 'vtt',
            ])->asText();

        return $response->text;
```
Which Laravel version do you use? Maybe, there's a difference because the original code passes a specific structure of parameters and file. The code in the PR is much more robust.

sixlive:
Closing in favor of https://github.com/prism-php/prism/pull/535

PR #514, opened 2025-07-25
fix: tool call and result chunk types in Gemini stream

kauffinger:
Adds handling for ChunkType::ToolCall and ChunkType::ToolResult in Gemini stream processing. Updates tests and fixtures to verify tool call and result chunk emission, including assertions for tool call arguments and tool result content.

## Breaking Changes
None


kauffinger:
I've just seen #452 exists, which would be fine with me, too. This PR just aligns to OAI, but no meta / other stuff yet.

PR #536, opened 2025-08-01
docs(media): update parameter names for fromStoragePath

skatecram:
## Description
The wrong parameter name is used in the docs.
In all `fromStoragePath` functions, change `disk` to `diskName` as used in the code.

Also remove any trailing whitespace at the ends of lines.

## Breaking Changes
-


sixlive (approved):
Thank you so much!

PR #560, opened 2025-08-15
feat(gemini): add image generation support

beefan:
## Description
Per https://github.com/prism-php/prism/issues/532, there is still opportunity to support more image generation models.

This PR tackles Gemini only as that is what I am most familiar with. It follows documentation [here](https://ai.google.dev/gemini-api/docs/image-generation) and [here](https://ai.google.dev/gemini-api/docs/imagen) to add support for:
- `gemini-2.0-flash-preview-image-generation` - an experimental, conversational style image generation model with image editing support
- `imagen-4.0-generate-001` family of models

I tried to follow [the PR which spearheaded image generation support](https://github.com/prism-php/prism/pull/428) as closely as possible for style, organization, and updating documentation. 

## Breaking Changes
- No breaking changes as far as I know

## Testing 
- Added `tests/Providers/Gemini/ImagesTest.php`, utilizing `FixtureResponse` to record initial responses from the actual API. 
- Manual testing


MordiSacks:
It looks like the production dos did not update this
https://prismphp.com/core-concepts/image-generation.html

PR #578, opened 2025-08-26
Feature/openrouter provider routing

mboisvertdupras:
## Summary

- Added support for OpenRouter provider-specific routing options including model ordering, fallback control, and provider filtering
- Implemented validation and filtering for all provider options to ensure only valid values are sent to the API

## Changes

- New `HandlesProviderOptions` trait with validation logic
- Integration across all OpenRouter handlers (Text, Stream, Structured)
- Refactored OpenRouter text tests to use `asText()` instead of `generate()` for consistency
- Added test suite for provider options validation
- Covers both valid option handling and invalid option filtering

There should be no breaking changes.


src/Providers/OpenRouter/Handlers/Structured.php
```diff
@@ -66,6 +66,7 @@ protected function sendRequest(Request $request): array
                         'schema' => $request->schema()->toArray(),
                     ],
                 ],
+                'provider' => $request->providerOptions('provider') ?? null,
```
anleac, line 69: NIT: Is there a reason for defaulting to null rather than omitting the field if not set? (Some APIs prefer missing keys over explicit nulls.)

anleac (approved):
LGTM 👍 🚀

sixlive (approved):

PR #581, opened 2025-08-27
feat(openAI): Add File storage interaction

bso-AL:
## Description

First of all thanks for this Package, TJ. I needed some extra functionality for a personal project, but thought maybe more people could make use of it - if you believe this fits within Prism's scope.

This PR adds the functionality of interacting with OpenAI's file storage. It implements the functionality of uploading a file, listing all files in your storage, deleting a file, retrieving a file's information and retrieving a file's content. 

I've tried to mimic the current structure as best as possible, but I was a little stumped with regards to making a ResponseBuilder as I see the other "types" have. I ended up making it into a provider specific concern that formats the Responses. 

I've tried to keep the Responses as generic as possible in order to future proof if other people want to implement the File storage interaction with other Providers other than OpenAI.

### DIFF
--- a/src/Providers/OpenAI/Handlers/File.php
+++ b/src/Providers/OpenAI/Handlers/File.php
@@ -30,25 +30,22 @@ class File
 
     public function uploadFile(): Response
     {
+
         if (! $this->fileName || ! $this->path || ! $this->disk) {
             throw new PrismException('File name, disk and path are required');
         }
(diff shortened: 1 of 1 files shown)


sixlive:
Since this is very specific to OpenAI, what if we use use it as `Prism::provider('openai')->uploadFile()` etc..

PR #601, opened 2025-09-04
New streaming output

sixlive:
## Description

... coming soon...

## Breaking Changes

LOL... streaming is completely different now.


Fludem:
Does this mean streaming isn't supported for OpenRouter at the moment? @sixlive 

The docs just say it's supported that's all

PR #610, opened 2025-09-10
feat(mcp): Adds mcp_servers client option support for Anthropic

Arkitecht:
## Description
Add support for mcp_servers client options for Anthropic (https://docs.anthropic.com/en/api/messages#body-mcp-servers)


sixlive (approved):
SICK! Thanks!!

PR #640, opened 2025-09-27
feat(openai): Multi-image edit support

sixlive:
## Description

Allows for multiple images to be sent for image edits.

For example, in the test we include the sunset background as well as the diamond and we get:

image

## Breaking Changes

- Image edits for OpenAI are no longer passed via `withProviderOptions` and are now passed via `Image` value objects in the `withPrompt`'s `additionalContent` parameter.


docs/providers/openai.md
```diff
@@ -265,6 +264,46 @@ $response = Prism::image()
 file_put_contents('edited-image.png', base64_decode($response->firstImage()->base64));
 ```
 
+#### Using Masks for Targeted Editing
+
+For precise control over which parts of the image to edit, use a mask image:
+
+```php
+$mask = fopen('tests/Fixtures/diamond-mask.png', 'r');
```
aimeos, line 272: Passing a resource isn't the best way as e.g. Gemini needs the mime type of the file too. So passing an Image value object would be a better way

sixlive: Great suggestion, I just made the update! (https://github.com/prism-php/prism/pull/640/commits/cf6b37b9817821df5eec5dcfe3b96e81068f67fe)

ProjektGopher (approved):
Enthusiastically excited for this improvement!

I've left a couple of comments, but this exactly what I was hoping for!

src/Providers/OpenAI/Handlers/Images.php
```diff
@@ -60,23 +62,32 @@ protected function sendRequest(Request $request): ClientResponse
...
-                $request->providerOptions('image'),
-            );
+        /** @var Image $image */
+        foreach ($request->additionalContent() as $index => $image) {
+            $this
+                ->client
+                ->attach(
+                    "image[{$index}]",
+                    $image->resource()
+                );
```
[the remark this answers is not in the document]
aimeos: When passing a Media value object, it can be extended to hold the file name too

PR #642, opened 2025-09-28
bug(gemini) Set default values for groundingSupport indices

aimeos:
## Description

Add default values for startIndex and endIndex in CitationMapper as they are not passed if startIndex is zero and endIndex is at the end of the string.

## Breaking Changes

No breaking changes, bugfix only


aimeos:
@sixlive This fix is rather important because Gemini API very often returns citations starting at the beginning and that causes an PHP error every time, rendering Gemini integration not very useful at the moment

PR #648, opened 2025-10-05
chore: Create .gitattributes

ijpatricio:
## Description

While going through some source dive, I realised that there are many unnecessary files and folders being exported by composer upon including the package as a dependency.

image


sixlive:
I dig this, we should probably keep the LICENSE file though.

ijpatricio:
Thanks! Yes! I then readjusted.

atm, this is the ignored files:

image

PR #655, opened 2025-10-11
[DRAFT] OPENAI Multiple tool call optimization with structured output support

SolStis86:
## Description

OPENAI Handler Optimization For Tool Calling 

This is just a drafting conceptual PR that enables tool calls with structured output via the Text handler.  

- It resolves the reasoning item request errors:

```
{
	"error": {
		"message": "Item 'rs_08a5a0af4c2676750068ea635cd7e08195b78474af4580f078' of type 'reasoning' was provided without its required following item.",
		"type": "invalid_request_error",
		"param": "input",
		"code": null
	}
}
```
- Adds configuration for parallel tool calls

Like I say - this is just drafting and requires tests writing etc but am using it locally and it works great

### DIFF
--- a/src/Text/Request.php
+++ b/src/Text/Request.php
@@ -43,7 +43,6 @@ class Request implements PrismRequest
         protected string|ToolChoice|null $toolChoice,
         array $providerOptions = [],
         protected array $providerTools = [],
-        protected bool $parallelToolCalls = true,
     ) {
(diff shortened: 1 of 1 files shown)


RhysLees:
@SolStis86 Anything i can do to help get this completed? also running into this issue

PR #675, opened 2025-10-19
docs(guidelines): update guidelines

pushpak1300:
This PR refactors the Boost guidelines for Prism. The earlier guidelines were around 4k tokens, which made them too large and often filled up the AI agent’s context. The new approach takes a more balanced route by fetching documentation at runtime to retrieve only the specific information needed.

PR #676, opened 2025-10-20
chore: compress core.blade.php

kauffinger:
In an earlier release we added the publishable boost guidelines.

These were far too long (my boost guidelines without are 450 lines, with prism they are over 1.2k lines). Claude even complains about the large context.

I've heavily compressed the file using claude as a quick fix. I'd still suggest curating a guide that *works well* in the future.


kauffinger:
Closing in favor of #675

PR #684, opened 2025-10-23
feat(api): Optional model parameter

aimeos:
## Description

The using() method currently requires the provider and a model name but there are providers which won't allow you to select the model or choose the current default model if no value is provided. This PR makes the $model parameter in using() optional and uses and empty string if no model name is passed.

## Breaking Changes

None


aimeos:
The formatting errors are not mine

sixlive (approved):

PR #687, opened 2025-10-23
fix: response containing ```json

sv-LayZ:
Updating the decodeObject method for the structured output to handle response who contain `'''json` at the beginning of the text.
The problem happened when we make a request with Claude 4.5 on AWS bedrock but fixing it on this lib instead of prism-php/bedrock to prevent it happening with an other provider in the future.


jstoone:
Absolutely incredible to see the contributions on here. Awesome work both of you. 🎉

sixlive (approved):

PR #696, opened 2025-10-25
feat(api) Use HasPrompts to pass system prompt to image requests

aimeos:
## Description

Allow system prompts for image requests and simplify/streamline code by using HasPromts concern. Some providers can use additional instructions, others will generate correct image content only without any additional instructions. Therefore, system instructions and prompts must be passed separately and used depending on the provider.

## Breaking Changes

None for users but the image request object requires a new parameter now.


aimeos:
The last PHPStan error seems to have something to do with one test but I don't know how to fix it.

PR #698, opened 2025-10-26
feat(stream/anthropic): Add Server Tool Use and Web Search Tool Results to Additional Content

roymckenzie:
## Description
In order to support returning server tool calls and web search tool results back to anthropic in multi-turn conversations, the developer should be able to access the server tool uses and web search tool results.

This PR adds the ability to handle `server_tool_use` and `web_search_tool_result` `content_block_start` events in the Anthropic Stream handler. It also handles server tool call input deltas.

This PR also updates the message stop event to add `server_tool_calls` and `web_search_tool_results` to the `additionalContent` param of the `StreamEndEvent` event so streamers can access the server tool calls and web search tool results for storing in the database and return to Anthropic on future turns.

This PR also updated the `AnthropicStreamState` to store the server tool calls (for appending new deltas) and the web search tool results for inclusion on the `StreamEndEvent` event.

This PR also updates the Anthropic `MessageMap` to check for `server_tool_calls` and `web_search_tool_results` to map them and add them to the `AssistantMessage` content.

The `web_search_tool_result` cannot be provided to Anthropic in a separate `ToolResultMessage` because Anthropic expects it to be on the `AssistantMessage`.

This PR could probably include additional abstraction, perhaps creating a `ServerToolCallEvent` and `WebSearchToolResultEvent` to match the user provided tools and their respective `ToolCallEvent` and `ToolResultEvent`.

I initially tried shoehorning this into the existing `ToolCall` and `ToolResult` classes but `type` parameters that would have required changes to global value objects and because Anthropic does not want the `web_search_tool_result` passed into a `ToolResultMessage` I decided to implement the solution in the above way.

I am open to making additional changes on this before it is merged.

@sixlive @ChrisB-TL would appreciate your thoughts.

Thank you.


sixlive (changes requested):
Can we add some tests?

roymckenzie:
@sixlive I have implemented tests for the new vars and methods in `AnthropicStreamState` class and for the changes in the Anthropic `MessageMap::mapAssistantMessage` method. There is also an Anthropic Stream test to check the `$additionalContent` param when using Anthropic 'web_search` provider tool to ensure `server_tool_calls` and `web_search_tool_results` are present.

There was an issue with Rector in the checks I could not resolve. On my local using php@8.2 I, of course, ran `composer test` and `composer format` and I did not run into any errors. But the Github check for formatting is failing.

If you can give me a hint as to why the check is failing I would be glad to resolve the issue.

roymckenzie:
@sixlive I like it. I have a pull request (#711) into your [provider-tool-events](https://github.com/prism-php/prism/tree/provider-tool-events) branch to implement the Anthropic side of streaming support for provider tools calls based on the new `ProviderToolEvent` class you created.

Note: I don't think your branch has been rebased with main so the checks failed for my PR because of Rector version mismatches.

If my PR into your new branch is accepted and you accept #707, I would amend this PR significantly to remove the logic for handling `server_tool_use` type `content_block_start` events in favor for the implementation based on your superior abstraction.

PR #723, opened 2025-11-06
Create Tool from Laravel MCP Tools

barryvdh:
## Description

This convert Laravel MCP Tools to Prism Tools, at least on a basic level as far as I can tell.

I added a RawSchema because it look likes Laravel uses the same Schema output already, so converting back and forth as I did in https://github.com/prism-php/prism/issues/722#issue-3596532839 seems a bit useless and losing data.

I added a check to add the default back in the response when not set.

You can call it like this:

```php
$response = Prism::text()
    ->using(Provider::Anthropic, 'claude-3-5-sonnet-latest')
    ->withMaxSteps(2)
    ->withPrompt('What is the weather like in Paris?')
    ->withTools([Tool::make(CurrentWeatherTool::class)])
    ->asText();
```

I think we can call probably do something similar with https://github.com/modelcontextprotocol/php-sdk/blob/main/src/Schema/Tool.php (Edit: but wait until we have a 1.x release)

The benefit from using `Tool::make()` is that we can use that for Prism Tool objects/strings, Laravel MCP strings and possible other (like the php-sdk), and just autowire the dependencies and then check the type.

I'm not really sure how the annotations work, or if Prism supports that. I left that out now.

Fixes https://github.com/prism-php/prism/issues/722

### DIFF
--- a/src/Tools/LaravelMcpTool.php
+++ b/src/Tools/LaravelMcpTool.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Prism\Prism\Tools;
 
 use Illuminate\Validation\ValidationException;


barryvdh:
I've updated this for yielding results (which is not supported in Prism I guess, so we just implode them as strings).

If Laravel MCP starts doing more (like images etc) we might need to handle that differently though.

This seems to work in a project we're working on with Laravel MCP as server, and Prism for the internal chatbot (so same tools, without calling the mcp server)

PR #724, opened 2025-11-07
fix: only process provider calls as provider calls

kauffinger:
Currently, normal tool calls also send the 'provider tool call' event for OAI.

This stops that from happening.


sixlive (approved):
Thanks!

PR #758, opened 2025-11-22
Fix missing thought_signature when using Prism in streaming mode with Gemini 3 Pro and tool calls

decaluwepatrick:
When using Prism in streaming mode with gemini 3 pro and tool calls, the following error happens:

HTTP request returned status code 400:
```
{
"error": {
"code": 400,
"message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:spider` , position 2. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.",
"status": "INVALID_ARGUMENT"
}
}
```

This PR implements a similar fix previously applied for text mode: https://github.com/prism-php/prism/pull/752

In a future PR, I plan to refactor the Gemini streaming implementation to reuse the same ToolCallMap logic, as it would allow reduce duplication.


sixlive (approved):
Sick! Thanks!!

PR #762, opened 2025-11-23
fix: gemini tool call fails if thought is included

chisomgbata:
Gemini tool call fails if you set thinking budget greater than 0, this set the inlclude thoughts to be true, now when prism what to process the tool call and create a tool call map some of the array items are not of function call but of text that has the thoughts of the model.

Fix filter out all element that the key is not functionCall


chisomgbata:
@sixlive no breaking changes or failing tests.

sixlive (approved):

PR #764, opened 2025-11-24
(feat) gemini: add support for new thinkingLevel

patrickbrouwers:
Gemini 3 pro introduced a new `thinkingLevel` config to replace the thinking budget: https://ai.google.dev/gemini-api/docs/thinking#thinking-levels

I've also added support for a complete `thinkingConfig` so we can also disable thought summaries, keeping the original `thinkingBudget` backwards compatible. 

`thinkingBudget` is also compatible with 3.0-pro, but they don't recommend using it.

### DIFF
--- a/src/Providers/Gemini/Handlers/Stream.php
+++ b/src/Providers/Gemini/Handlers/Stream.php
@@ -449,6 +449,22 @@ class Stream
             $tools = ['function_declarations' => ToolMap::map($request->tools())];
         }
 
+        $thinkingConfig = $providerOptions['thinkingConfig'] ?? null;
+
+        if (isset($providerOptions['thinkingBudget'])) {
+            $thinkingConfig = [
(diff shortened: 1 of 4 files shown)


cutlope:
> I'm personally not a big fan of having the aliases, but I included it to not introduce a breaking change. Setting `'includeThoughts' => true` was already the behaviour before this PR, so was kept to not introduce a breaking change.
> 
> I personally think it's much better to pass a thinkingConfig array and handle includeThoughts explicitly; that's how I'm doing it myself:
> 
> ```
> 'thinkingConfig' => [
>      'thinkingBudget' => 1024,
>      'includeThoughts' => true,
> ]
> ```

Yeah apologies, it wasn't this pr that introduced it. In fact, to my surprise, it was my own previous pr #729 that added it to the `Text.php` file
That pr aligned it with the `Stream.php` changes that were introduced in #601 and that seemed to have the default `'includeThoughts' => true`

PR #775, opened 2025-11-29
fix(gemini): only include includeThoughts when explicitly set

cutlope:
## Description
After PR #764 setting `thinkingBudget` or `thinkingLevel` would automatically set `includeThoughts` to true. This change makes `includeThoughts` independent from thinking budget/level settings - it will only be included in the API request when explicitly set by the user.


patrickbrouwers:
As mentioned in #764 this change would actually be a breaking change for people having used `thinkingBudget` and depending on thoughts being included. I don't think most people use it, so probably not a lot of harm in changing the default and making it explicit (which I personally think would have been the best option orginally too)

sixlive (approved):

PR #780, opened 2025-12-03
Feature: Implements TTS for ElevenLabs

philipsorensen:
## Description
Adds TTS for Elevenlabs. Took inspiration from the Groq code and tested the feature in my project. 

## Breaking Changes
Shouldn't break anything.


sixlive:
Yooo! Sick! Can you add some tests?

PR #806, opened 2025-12-22
Feat: Add support for client-executed tools

vinitkadam03:
### Summary
This PR introduces support for **client-executed tools** (also referred to as "deferred tools") - tools that are defined without a handler function and are intended to be executed by the client/caller rather than by Prism.

### Motivation
Client-executed tools enable scenarios where tool execution must happen on the client side, such as:
- UI interactions requiring client-side execution
- Any tool where the server should not (or cannot) execute the logic

#### Core Implementation

**`src/Tool.php`**
- Added `isClientExecuted()` method that returns `true` when no handler function is defined (`$this->fn === null`)

```php
public function isClientExecuted(): bool
{
    return $this->fn === null;
}
```

**`src/Concerns/CallsTools.php`**
- Modified `callTools()` to filter out client-executed tools from execution
- Added `hasDeferredTools()` method to check if any tool calls are for client-executed tools

**Behavior:**
- Client-executed tools are skipped during tool execution
- Server-executed tools in the same request are still executed normally
- When client-executed tools are detected, execution stops and control is returned to the caller
- The LLM is not called for the next turn, allowing the client to execute the tool and continue the conversation
- Response/stream ends with FinishReason::ToolCalls

### Usage Example

```php
use Prism\Prism\Facades\Prism;
use Prism\Prism\Facades\Tool;

// Define a client-executed tool (no `using()` call)
$clientTool = Tool::as('client_tool')
    ->for('A tool that executes on the client')
    ->withStringParameter('input', 'Input parameter');
    // Note: No ->using() call!

```

## Breaking Changes
None. This is a backward-compatible addition. Existing tools with handlers continue to work exactly as before.

### DIFF
--- a/docs/core-concepts/tools-function-calling.md
+++ b/docs/core-concepts/tools-function-calling.md
@@ -305,6 +305,56 @@ use Prism\Prism\Facades\Tool;
 $tool = Tool::make(CurrentWeatherTool::class);
 ```
 
+## Client-Executed Tools
+
+Sometimes you need tools that are executed by the client (e.g., frontend application) rather than on the server. Client-executed tools are defined without a handler function - simply omit the `using()` call:
+
+```php
+use Prism\Prism\Facades\Tool;
+
+$clientTool = Tool::as('browser_action')
+    ->for('Perform an action in the user\'s browser')
+    ->withStringParameter('action', 'The action to perform');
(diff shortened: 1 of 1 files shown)


vinitkadam03:
@sixlive will update and add step finish events once #807 is reviewed and merged.

PR #809, opened 2025-12-23
Only send max_ouput_tokens if set

### LINKED ISSUE #799: OpenAI: max_output_tokens is included by default
According to the [docs](https://platform.openai.com/docs/api-reference/responses/create), the field is optional. Prism seems to add it regardless of the field being set or not.
The correct behavior would be not including the field at all by default.

sixlive:
## Description

Closes #799


TonsiTT:
@sixlive the issue is still there due to this line
`protected ?int $maxTokens = 2048;`
https://github.com/prism-php/prism/blob/c1e17ae2e892c3b0f6db9c61f2f33e847b373d76/src/Concerns/ConfiguresModels.php#L9C1-L9C38

sixlive:
I just tagged the fix! https://github.com/prism-php/prism/releases/tag/v0.99.6

PR #813, opened 2025-12-28
feat(openrouter): Improve error handling by extracting messages from metadata.raw.

pocketarc:
The OpenRouter provider only looked at `error.message`, which always yielded an unhelpful "Provider returned error" exception message. By looking at the `error.message` that came from the provider, we can get the _real_ error message, which can be far more helpful.

### Before

> OpenRouter Bad Request: Provider returned error

### After

> OpenRouter Bad Request: Invalid schema for response_format 'foo': In context=(), 'required' is required to be supplied and to be an array including every key in properties. Missing 'bar'.


sixlive (approved):
Sick! Thanks!!

PR #824, opened 2026-01-13
feat(openrouter): Include raw API response context in structured decoding exceptions.

pocketarc:
When structured decoding fails, the exception message was unhelpful; it would show just "Received: " with nothing after it.

By catching the exception in the OpenRouter handler (where we still have access to the raw API response) and re-throwing with context, we can now see exactly what the provider returned.

### Before

> Structured object could not be decoded. Received: 

### After

> Structured object could not be decoded. Received: 
> Model: anthropic/claude-3.5-sonnet
> Finish reason: stop
> Raw choices: [{"message": {"content": null, "role": "assistant"}}]

### DIFF
--- a/src/Providers/OpenRouter/Handlers/Structured.php
+++ b/src/Providers/OpenRouter/Handlers/Structured.php
@@ -6,6 +6,7 @@ namespace Prism\Prism\Providers\OpenRouter\Handlers;
 
 use Illuminate\Http\Client\PendingRequest;
 use Prism\Prism\Exceptions\PrismException;
+use Prism\Prism\Exceptions\PrismStructuredDecodingException;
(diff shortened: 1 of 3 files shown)


pocketarc:
Same! Or at least have the truncation be reasonable, like 4KB of text or something - you're still protected against ridiculously-long responses but you're certainly going to be able to see the full error message at once.

Also: I didn't get to see your talk at PHPTek but I was at the Prompt sponsor table - we were the ones that were there hiring! I've been raising PRs whenever I see anything that I've had to deal with when working with OpenRouter, but if there's anything you'd like me to do differently for this project, let me know! And if there's ever anything lying around that you'd like someone to take care of I'd be happy to look into it.

PR #825, opened 2026-01-13
fix(openrouter): Handle null id/model in API responses.

pocketarc:
OpenRouter can sometimes return responses without `id` or `model` fields in the response body. The handlers passed these values directly to `Meta::__construct()` without defaults, causing a TypeError when they were null.

### Before

```
TypeError: Prism\Prism\ValueObjects\Meta::__construct(): Argument #1 ($id) must be of type string, null given
```

### After

The response is handled, with `id` defaulting to an empty string and `model` falling back to the requested model from the original request.


sixlive (approved):
Good catch! Thanks!

PR #831, opened 2026-01-15
feat(openrouter): Add video support with correct video_url format.

pocketarc:
The OpenRouter provider had video support, but the payload format was incorrect. It was using `input_video` with a `data`/`format` structure, but OpenRouter's API expects `video_url` with a `url` field (either a direct URL or a base64 data URL).

### Before

```php
['type' => 'input_video', 'input_video' => ['data' => '...', 'format' => 'mp4']]
```

### After

```php
['type' => 'video_url', 'video_url' => ['url' => 'data:video/mp4;base64,...']]
```

I've updated the mapper to match the [OpenRouter video documentation](https://openrouter.ai/docs/guides/overview/multimodal/videos), added documentation, and included tests. Also added support for passing video URLs directly (e.g., YouTube links for Gemini models).

### DIFF
--- a/src/Providers/OpenRouter/Maps/VideoMapper.php
+++ b/src/Providers/OpenRouter/Maps/VideoMapper.php
@@ -28,8 +28,7 @@ class VideoMapper extends ProviderMediaMapper
             ];
         }
 
-        $mimeType = $this->media->mimeType() ?? 'video/mp4';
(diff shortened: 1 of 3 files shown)


sixlive (approved):
Sick! Thanks!

PR #884, opened 2026-01-31
Fix: Ensure step messages represent input only

vinitkadam03:
## Description

Updated **text handlers** and **stream handlers** across all providers where step messages contained both input and output messages. 

**In text handlers:** The assistant message (response) was being added to the request **before** `addStep()` was called, which caused each step's `messages` array to include the assistant's response alongside the input messages.

**In stream handlers:** The assistant and tool result messages were being added to the request **before** the `StepFinishEvent` was dispatched, which caused the same issue where each step's messages would include the output alongside the input.

### The Problem

In multi-step tool call scenarios:
- Step N's `messages` was including the assistant message (output) that was generated during that step
- Ideally `step->messages` should represent only the **input** to that step
- The output is already captured in other step properties (`text`, `toolCalls`, `toolResults`, etc.)

> **Note:** This bug can cause issues with observability/telemetry systems. If you're storing the steps returned by `asText()` or collected via streaming for logging, tracing, or analytics, the incorrect message structure would show the assistant's output duplicated - once in `step->messages` and again in `step->text`/`step->toolCalls`. This makes it difficult to accurately reconstruct the conversation flow and could lead to confusing traces. Additionally, it becomes tricky to determine what input was actually sent to the LLM at each step, as users have to cross-reference other attributes like `text`, `toolCalls`, `toolResults`, etc. to distinguish between input and output within the messages array.

### The Fix

**Text Handlers:** Moved the assistant message addition to **after** `addStep()` in all provider text handlers.

**Stream Handlers:** Moved the assistant and tool result message additions to **after** yielding `StepFinishEvent` in all provider stream handlers.

Affected providers:
- `Anthropic`, `OpenAI`, `OpenRouter`, `DeepSeek`, `Gemini`, `Groq`, `Mistral`, `Ollama`, `XAI`

Now in multi-step scenarios:
- Step 1: `messages` = `[user message]` (input only)
- Step 2: `messages` = `[user message, assistant message from step 1, tool results]` (input only)
- And so on...

### Tests Updated

Added assertions across all provider test files to verify:
- The assistant message from step N correctly appears in step N+1's input messages
- Each step's messages only contain the input to that step

### Text Handlers

`step->messages` will no longer include the assistant message generated during that step - only the input messages to that step. If you were accessing the assistant message via `$response->steps[N]->messages`, you should now use `$response->steps[N]->text` or `$response->steps[N]->toolCalls` instead.

**No breaking change** for `$response->messages->last()` - it still returns the final assistant message for conversation continuation.

### Stream Handlers

When the `StepFinishEvent` is dispatched after tool results are executed, the request messages will no longer contain the assistant and tool result messages at that point. These messages are now added to the request **after** the `StepFinishEvent` is yielded. If you were accessing `$request->messages()` in a `StepFinishEvent` listener expecting to see the assistant response and tool results, they will no longer be present at that time.

### DIFF
--- a/src/Providers/Anthropic/Handlers/Stream.php
+++ b/src/Providers/Anthropic/Handlers/Stream.php
@@ -503,6 +503,13 @@ class Stream
 
         // Add messages to request for next turn
         if ($toolResults !== []) {
+            // Emit step finish after tool calls
+            $this->state->markStepFinished();
+            yield new StepFinishEvent(
+                id: EventID::generate(),
+                timestamp: time()
+            );
+
             $request->addMessage(new AssistantMessage(
                 content: $this->state->currentText(),
                 toolCalls: $toolCalls,
@@ -514,13 +521,6 @@ class Stream
 
             $request->addMessage(new ToolResultMessage($toolResults));
             $request->resetToolChoice();
-
-            // Emit step finish after tool calls
-            $this->state->markStepFinished();
-            yield new StepFinishEvent(
-                id: EventID::generate(),
-                timestamp: time()
-            );
 
             // Continue streaming if within step limit
             $depth++;
--- a/src/Providers/DeepSeek/Handlers/Stream.php
+++ b/src/Providers/DeepSeek/Handlers/Stream.php
@@ -383,15 +383,15 @@ class Stream
         $toolResults = [];
         yield from $this->callToolsAndYieldEvents($request->tools(), $mappedToolCalls, $this->state->messageId(), $toolResults);
 
-        $request->addMessage(new AssistantMessage($text, $mappedToolCalls));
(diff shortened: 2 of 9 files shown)


vinitkadam03:
I just checked the stream handler, and when the finish reason is stop, the assistant message is not added to the request before dispatching the step-finish and stream-end events.

I think the same behavior should apply to the text request state as well—the messages should represent only the input messages sent to the LLM to generate the output for that step.

I also checked what happens when the finish reason is tool_calls in the stream handlers. In that case, the assistant message is added before dispatching the step-finish event.
This feels a bit inconsistent. Ideally, the step-finish event should be dispatched before adding the assistant message and tool results to the request in stream handlers as well.

sixlive (approved):

PR #903, opened 2026-02-11
fix: prevent duplicate reasoning-end events in Ollama stream

bishwajitcadhikary:
# Description
This PR fixes an issue where the Ollama stream handler would repeatedly emit `ThinkingCompleteEvent` (mapping to `reasoning-end`) for every chunk after a reasoning block had finished.

**The Problem:**
The `Stream` handler checks `if ($this->state->hasThinkingStarted())` to determine if it should emit a `ThinkingCompleteEvent`. However, the `StreamState` class lacked a method to reset `thinkingStarted` to `false`. As a result, once asking for reasoning began, the flag remained `true` for the duration of the stream, causing `ThinkingCompleteEvent` to be yielded for every subsequent data chunk that didn't contain a thinking delta.

**The Fix:**
1.  Added `markThinkingCompleted()` to `StreamState` to allow resetting the `thinkingStarted` flag.
2.  Updated `Providers\Ollama\Handlers\Stream` to call `$this->state->markThinkingCompleted()` immediately before yielding `ThinkingCompleteEvent`.

This ensures the event is emitted exactly once per reasoning block, preventing frontend errors (such as `AI_UIMessageStreamError: Received reasoning-end for missing reasoning part`) caused by duplicate end events.

# Breaking Changes
None. This is a bug fix that adds a method to `StreamState` and corrects internal state management logic.


sixlive:
I'd like to merge this but I need you to rebase against origin/main. I tried to do it but I can't push to your branch.

PR #937, opened 2026-03-02
fix: sanitize control characters in ToolCall::arguments() before json_decode

### LINKED ISSUE #936: JsonException: Control character error in ToolCall::arguments() with DeepSeek streaming
## Bug Description

`ToolCall::arguments()` throws `JsonException: Control character error, possibly incorrectly encoded` when DeepSeek's streaming API returns tool call argument JSON containing raw control characters (bytes 0x00-0x1F).

This has occurred **11 times in production** over a short period.

## Stack Trace

```
JsonException: Control character error, possibly incorrectly encoded
  at src/ValueObjects/ToolCall.php:39
  at src/Concerns/CallsTools.php:136
  at src/Concerns/CallsTools.php:118
  at src/Concerns/CallsTools.php:53
  at src/Concerns/CallsTools.php:34
  at src/Providers/DeepSeek/Handlers/Text.php:68
```

## Root Cause

In `ToolCall::arguments()`, the arguments string is passed directly to `json_decode()` with `JSON_THROW_ON_ERROR`:

```php
return json_decode(
    $arguments,
    true,
    flags: JSON_THROW_ON_ERROR
);
```

When DeepSeek streams tool call arguments via deltas, the concatenated string can contain raw control characters that are invalid in JSON.

## Comparison with OpenRouter

The OpenRouter streaming handler already has a try-catch fallback for this exact scenario:

```php
try {
    $parsedArguments = json_decode($arguments, true, flags: JSON_THROW_ON_ERROR);
    $arguments = $parsedArguments;
} catch (Throwable) {
    $arguments = ['raw' => $arguments];
}
```

The DeepSeek handler has no such protection.

## Suggested Fix

Sanitize control characters before `json_decode` in `ToolCall::arguments()`:

```php
// Sanitize control characters that some providers may include
$arguments = preg_replace('/[\x00-\x1F\x7F]/', '', $this->arguments);

return json_decode(
    $arguments,
    true,
    flags: JSON_THROW_ON_ERROR
);
```

This is safe because raw control characters (0x00-0x1F, 0x7F) are never valid in JSON — they must be escaped as `\uXXXX` sequences per RFC 8259.

## Environment

- **Prism version**: v0.99.19
- **Provider**: DeepSeek (streaming)
- **PHP**: 8.4
- **Laravel**: 12

ruttydm:
## Summary

- Sanitize raw control characters (0x00-0x1F, 0x7F) from tool call argument strings before `json_decode` in `ToolCall::arguments()`
- These bytes are invalid in JSON per RFC 8259 and cause `JsonException: Control character error, possibly incorrectly encoded`
- Observed 11 times in production with the DeepSeek streaming provider

## Context

Some providers (notably DeepSeek) occasionally include raw control characters in streamed tool call argument JSON. The OpenRouter handler already has defensive handling for malformed arguments — this adds similar protection at the `ToolCall` level so all providers benefit.

Fixes #936

### DIFF
--- a/src/ValueObjects/ToolCall.php
+++ b/src/ValueObjects/ToolCall.php
@@ -34,7 +34,10 @@ class ToolCall implements Arrayable
                 return [];
             }
 
-            $arguments = $this->arguments;
+            // Sanitize control characters that some providers (e.g. DeepSeek) may include
+            // in streamed tool call arguments. Raw 0x00-0x1F / 0x7F bytes are never valid
+            // in JSON (RFC 8259) and cause json_decode to throw "Control character error".
+            $arguments = preg_replace('/[\x00-\x1F\x7F]/', '', $this->arguments);
 
             return json_decode(
                 $arguments,
--- a/src/ValueObjects/ToolCall.php
+++ b/src/ValueObjects/ToolCall.php
@@ -40,7 +40,7 @@ class ToolCall implements Arrayable
             $arguments = preg_replace('/[\x00-\x1F\x7F]/', '', $this->arguments);
 
             return json_decode(
-                $arguments,
+                (string) $arguments,
                 true,
                 flags: JSON_THROW_ON_ERROR
             );


sixlive (approved):

sixlive (changes requested):
Can you address CI errors please

ruttydm:
Rebased on latest main (up to #970). CI doesn't seem to trigger for fork PRs — could a maintainer approve the workflow run?

PR #960, opened 2026-03-17
feat(xai): Add XAI image generation handler and tests

timothymarois:
This pull request introduces image generation support for the XAI provider, enabling the handling and mapping of image generation requests and responses. The main additions are a new `Images` handler, a request mapping utility, and integration of these into the provider interface.

**Image generation support:**

* Added the `Images` handler class in `src/Providers/XAI/Handlers/Images.php` to process image generation requests, validate responses, extract generated images, and build structured responses.
* Added the `ImageRequestMap` utility in `src/Providers/XAI/Maps/ImageRequestMap.php` to map image request parameters for the XAI API, including support for additional provider options.

**Provider integration:**

* Registered the new image request and response types, and imported the `Images` handler in `src/Providers/XAI/XAI.php`.
* Added the `images()` method to the XAI provider class to handle image generation requests using the new handler.Introduce image generation support for the XAI provider: add Images handler (sends POST to images/generations, validates response, extracts GeneratedImage objects, and builds a Response via ResponseBuilder) and ImageRequestMap (maps Request to provider payload and forwards supported and additional provider options). Wire the new handler into XAI by adding an images() method. Add comprehensive tests for URL and base64 responses, provider-specific options, multiple images, usage, and meta handling.

## Description
docs say it supports xai image generation, but it actually returns an exception when trying to use xai. 

## Breaking Changes
None

### DIFF
--- a/src/Providers/XAI/Handlers/Images.php
+++ b/src/Providers/XAI/Handlers/Images.php
@@ -1,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Prism\Prism\Providers\XAI\Handlers;
+
+use Illuminate\Http\Client\PendingRequest;
+use Illuminate\Http\Client\Response as ClientResponse;
+use Prism\Prism\Images\Request;
+use Prism\Prism\Images\Response;
+use Prism\Prism\Images\ResponseBuilder;
+use Prism\Prism\Providers\XAI\Concerns\ValidatesResponses;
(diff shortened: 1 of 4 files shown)


mssayari:
It seems that XAI is also providing the mime_type. If it exists in the response, it would be great if we could also map it.


The Documentation URL:
[https://docs.x.ai/developers/rest-api-reference/inference/images#image-generation](url)

Issue #520, opened 2025-07-29
Cloudflare Workers AI


babul:
Would love to see support for Cloudflare Workers AI models.

kyranb:
Just for reference, here are the docs here: https://developers.cloudflare.com/workers-ai/

PR #964, opened 2026-03-18
Fix xAI driver compatibility with OpenAI-compatible endpoints

meirdick:
## Problem

The xAI driver doesn't work with OpenAI-compatible `/chat/completions` endpoints like Cloudflare Workers AI's `/compat` endpoint. Two issues:

### 1. User messages always use array content format

`MessageMap` serializes all user messages as:
```json
{"role": "user", "content": [{"type": "text", "text": "Hello"}]}
```

Some OpenAI-compatible endpoints only accept string content:
```json
{"role": "user", "content": "Hello"}
```

The array format causes the model to receive garbled input and respond with nonsense, or reject the request with: `"Type mismatch of '/messages/0/content', 'array' not in 'string'"`.

### 2. Structured handler crashes when content is returned as object

The structured handler expects `choices.0.message.content` to be a string (per the OpenAI spec), but some providers return it as a parsed object/array:

```json
// Expected (xAI/OpenAI):
{"content": "{\"title\":\"Inception\"}", "parsed": {"title": "Inception"}}

// Actual (Workers AI):
{"content": {"title": "Inception"}}
```

This crashes on `new AssistantMessage($content)` with: `TypeError: Argument #1 ($content) must be of type string, array given`.

## Fix

1. **MessageMap**: Use plain string content when no images are present, array format only when images exist. Both formats are valid per the OpenAI spec.

2. **Structured handler**: Normalize array content to a JSON string via `json_encode()`, and use it as the parsed data when no `parsed` field is present.

Both changes are backward-compatible with xAI's native API — verified by all existing tests passing.

## Testing

- All 49 existing xAI tests pass
- Added new test: `it('handles content returned as object instead of string')` covering the Workers AI response format
- Updated `MessageMapTest` to expect string content for text-only messages

## Context

Relates to #520. The xAI driver is commonly used with Cloudflare Workers AI (via the `/compat` endpoint) because it sends to `/chat/completions` rather than `/responses`. These two fixes make that combination work reliably for text generation and structured output.

### DIFF
--- a/src/Providers/XAI/Handlers/Structured.php
+++ b/src/Providers/XAI/Handlers/Structured.php
@@ -41,8 +41,14 @@ class Structured
 
         $this->handleRefusal(data_get($data, 'choices.0.message', []));
 
-        $content = data_get($data, 'choices.0.message.content') ?? '';
+        $rawContent = data_get($data, 'choices.0.message.content') ?? '';
         $parsed = data_get($data, 'choices.0.message.parsed');
+
+        // Some OpenAI-compatible providers (e.g. Cloudflare Workers AI) return
+        // content as a parsed object/array instead of a JSON string. Normalize
+        // to string for AssistantMessage and extract structured data.
+        $content = is_array($rawContent) ? json_encode($rawContent) : $rawContent;
+        $parsed ??= is_array($rawContent) ? $rawContent : null;
 
         $responseMessage = new AssistantMessage($content);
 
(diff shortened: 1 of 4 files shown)


SergkeiM:
Hi @meirdick as I understood current workaround is your package: [/prism-workers-ai](https://github.com/meirdick/prism-workers-ai) (thank you)

But is definetly nice to have this built-in in prism

meirdick:
Hi @SergkeiM, yes that is correct. The package provides the full driver for complete worker-ai compatibility. It also makes for a clean automated implementation via this skill:
https://github.com/meirdick/laravel-cloudflare-ai-gateway

SergkeiM:
Hi @sixlive what are your thoughts of adding CloudFlare support (merging this PR)?

meirdick:
> Hi @sixlive what are your thoughts of adding CloudFlare support (merging this PR)?

@SergkeiM - Looks like Laravel/AI is heading in the direction of adding direct gateway integration (first party and otherwise).  I made a PR there, but see comment there: 
https://github.com/laravel/ai/pull/405

PR #993, opened 2026-03-28
feat(openrouter): support cacheTtl provider option for extended cache TTL

kargnas:
OpenRouter supports extended cache TTL via the ttl field: https://openrouter.ai/docs/guides/best-practices/prompt-caching

Before (5 min default, no way to extend):
```
  $message->withProviderOptions(['cacheType' => 'ephemeral']);
  // sends: "cache_control": {"type": "ephemeral"}
```

After (1h TTL):
```
  $message->withProviderOptions(['cacheType' => 'ephemeral', 'cacheTtl' => '1h']);
  // sends: "cache_control": {"type": "ephemeral", "ttl": "1h"}
```

Without cacheTtl, behavior is unchanged (no ttl key sent).


kargnas:
@sixlive busy?

PR #994, opened 2026-03-29
fix(gemini,groq): include finish reason in unhandled finish reason exceptions

datashaman:
## Description

Enriches the default `PrismException` in `Gemini\Handlers\Text`, `Gemini\Handlers\Structured`, and `Groq\Handlers\Text` to include both the mapped `FinishReason` enum value and the raw provider finish reason string — matching the approach already used in OpenAI handlers (PR #941).

Previously threw a generic `"Gemini: unhandled finish reason"` with no indication of what the actual reason was (e.g. `SAFETY`, `LANGUAGE`, etc.). Now throws `'Gemini: unhandled finish reason "content-filter" (raw: SAFETY)'`.

Relates to #205.

**Changes:**
- `src/Providers/Gemini/Handlers/Text.php` — enriched default exception with sprintf
- `src/Providers/Gemini/Handlers/Structured.php` — same
- `src/Providers/Groq/Handlers/Text.php` — same
- Added Pest tests for Gemini Text and Structured content filter exceptions
- Added test fixtures simulating SAFETY finish reason responses

## Breaking Changes

None — only changes exception message strings in error paths.

### DIFF
--- a/src/Providers/Gemini/Handlers/Structured.php
+++ b/src/Providers/Gemini/Handlers/Structured.php
@@ -68,7 +68,11 @@ class Structured
         return match ($finishReason) {
             FinishReason::ToolCalls => $this->handleToolCalls($data, $request),
             FinishReason::Stop, FinishReason::Length => $this->handleStop($data, $request, $finishReason),
-            default => throw new PrismException('Gemini: unhandled finish reason'),
+            default => throw new PrismException(sprintf(
+                'Gemini: unhandled finish reason "%s" (raw: %s)',
+                $finishReason->value,
+                data_get($data, 'candidates.0.finishReason', 'unknown'),
+            )),
         };
     }
 
--- a/src/Providers/Gemini/Handlers/Text.php
+++ b/src/Providers/Gemini/Handlers/Text.php
@@ -59,7 +59,11 @@ class Text
         return match ($finishReason) {
             FinishReason::ToolCalls => $this->handleToolCalls($data, $request),
             FinishReason::Stop, FinishReason::Length => $this->handleStop($data, $request, $finishReason),
-            default => throw new PrismException('Gemini: unhandled finish reason'),
+            default => throw new PrismException(sprintf(
+                'Gemini: unhandled finish reason "%s" (raw: %s)',
+                $finishReason->value,
+                data_get($data, 'candidates.0.finishReason', 'unknown'),
+            )),
         };
     }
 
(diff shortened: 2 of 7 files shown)


datashaman:
## Additional context: unhandled finish reasons

While investigating this fix, I catalogued the complete set of finish reasons each provider can return vs what's currently handled.

### Gemini

Gemini's `FinishReasonMap` currently maps 10 values, but Google's API defines at least 20+. These newer values are **not in the map** (they fall through to `default => FinishReason::Other`, then hit the exception this PR improves):

| Value | Notes |
|---|---|
| `IMAGE_SAFETY` | Image generation safety violation |
| `IMAGE_PROHIBITED_CONTENT` | Image generation prohibited content |
| `IMAGE_RECITATION` | Image generation recitation |
| `IMAGE_OTHER` | Image generation misc |
| `NO_IMAGE` | Expected image but none produced |
| `UNEXPECTED_TOOL_CALL` | Tool call generated but no tools enabled |
| `TOO_MANY_TOOL_CALLS` | Too many consecutive tool calls (Google AI only) |
| `MISSING_THOUGHT_SIGNATURE` | Thought signature missing (Google AI only) |
| `MALFORMED_RESPONSE` | Malformed response (Google AI only) |
| `MODEL_ARMOR` | Blocked by Model Armor (Vertex AI only) |

Additionally, 8 values that **are** in the `FinishReasonMap` (mapping to `ContentFilter` or `Other`) still hit the default exception branch in the handler because only `Stop`, `Length`, and `ToolCalls` are matched:

`SAFETY`, `RECITATION`, `BLOCKLIST`, `PROHIBITED_CONTENT`, `SPII`, `MALFORMED_FUNCTION_CALL`, `LANGUAGE`, `OTHER`, `FINISH_REASON_UNSPECIFIED`

### Groq

Groq only has 4 finish reasons (`stop`, `length`, `tool_calls`, `function_call`). The first three are handled. `function_call` (deprecated) is not in the `FinishReasonMap`.

---

These are potential follow-up items — this PR just ensures the exception messages are useful for debugging when any of them are encountered.

Sources: [Google AI REST v1](https://ai.google.dev/api/rest/v1/GenerateContentResponse#FinishReason), [Vertex AI REST v1](https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/GenerateContentResponse), [Groq API reference](https://console.groq.com/docs/api-reference#chat-create)

mnapoli:
That would be very helpful to have!

PR #996, opened 2026-03-30
fix: handle unknown finish reasons gracefully

cainydev:
## Summary
- All provider handlers crashed with `PrismException` on unrecognized finish reasons (mapped to `FinishReason::Unknown`/`Other` by `FinishReasonMap`)
- Routes unknown finish reasons to `handleStop()` across all 14 affected handlers in 10 providers, returning content instead of crashing
- Preserves the `FinishReason` enum value in the response for caller inspection

## Test plan
- [x] All 1463 tests pass
- [x] Updated OpenAI Text/Structured tests to verify graceful handling instead of expecting exceptions

### DIFF
--- a/src/Providers/Anthropic/Handlers/Structured.php
+++ b/src/Providers/Anthropic/Handlers/Structured.php
@@ -11,7 +11,6 @@ use InvalidArgumentException;
 use Prism\Prism\Concerns\CallsTools;
 use Prism\Prism\Contracts\PrismRequest;
 use Prism\Prism\Enums\FinishReason;
-use Prism\Prism\Exceptions\PrismException;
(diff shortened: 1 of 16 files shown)


markritterman:
Hey @cainydev, thanks for tackling this! We just hit this exact failure with `qwen3-coder-next` via OpenRouter in structured-output mode — the handler threw `OpenRouter: unknown finish reason` on responses qwen emits with a `finish_reason` Prism doesn't map. Your `default => handleStop()` fix would have saved us from shipping a local workaround.

We'll track this PR and upgrade as soon as it lands. If there's anything useful I can do to help it along (testing, feedback on the approach), happy to. Thanks again for the fix! 🙌

PR #999, opened 2026-03-31
feat(OpenAI): Enable structured output support for OpenAI gpt-5.4 mini and nano models

aivis:
## Description
This is a follow-up for https://github.com/prism-php/prism/pull/947

A few days after OpenAI released 5.4 model variations with structured output support:
- https://developers.openai.com/api/docs/models/gpt-5.4-mini
- https://developers.openai.com/api/docs/models/gpt-5.4-nano

## Breaking Changes
No breaking changes.


aivis:
However, this might be a better candidate to fix this https://github.com/prism-php/prism/pull/991

PR #1000, opened 2026-04-02
fix(Gemini): Add `Usage` to `StepFinishEvent`

Plytas:
## Description

When using `->withTools()->asStream()` in Gemini, we're missing token usage on the intermediate LLM call:

```
  User Message
       │
       ▼
  ┌─────────┐    usage: prompt=100, comp=200, thought=50
  │ LLM #1  │──→ StepFinishEvent  ❌ no usage field
  └────┬────┘
       │ tool calls
       ▼
  ┌──────────┐
  │ Tool Exe │
  └────┬─────┘
       │ tool results
       ▼
  ┌─────────┐    usage: prompt=200, comp=400, thought=0
  │ LLM #2  │──→ StepFinishEvent  ❌ no usage field
  └────┬────┘
       │ final text
       ▼
  StreamEndEvent
    usage: prompt=200, comp=400  ← ❌ only LLM #2
```

There is a code block in Gemini `Stream` handler that supposedly updates usage to include previous usage: https://github.com/prism-php/prism/blob/5d6cc65b80b19cf3f22744703ac0c727b68cdca8/src/Providers/Gemini/Handlers/Stream.php#L359-L367

However, this is effectively dead code as `StreamEndEvent` is fired before this and there's no way to retrieve this updated usage. There are no tests that hit this block.

## Proposed solution

I propose to expose `Usage` via `StepFinishEvent`. This PR only adds it to Gemini provider (as that's what I have access to), but I'd be able to add it to all instances of `StepFinishEvent` if needed.

### DIFF
--- a/src/Providers/Gemini/Handlers/Stream.php
+++ b/src/Providers/Gemini/Handlers/Stream.php
@@ -241,7 +241,8 @@ class Stream
         $this->state->markStepFinished();
         yield new StepFinishEvent(
             id: EventID::generate(),
-            timestamp: time()
+            timestamp: time(),
+            usage: $this->state->usage(),
         );
 
         yield $this->emitStreamEndEvent();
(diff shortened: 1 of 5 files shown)


patrickbrouwers:
This would be super nice for all providers I think

PR #1001, opened 2026-04-02
fix(Media): normalize mimeType for audio/wav

### LINKED ISSUE #981: .wav files not working in OpenAI because of audio/x-wav instead of audio/wav mimeType
When I try to transcribe a .wav file with OpenAI, I get:

>  OpenAI Error [400]: invalid_request_error - Audio file might be corrupted or unsupported

If I override the mimeType to `audio/wav`, it works correctly.

```php
$response = Prism::audio()
    ->using('openai', 'gpt-4o-transcribe')
    ->withInput(Audio::fromLocalPath(path: '/path/to/file.wav', mimeType: 'audio/wav'))
    ->asText();
```

This is because the Media.php detect it like this:

```
        if (! $mimeType && ! ($mimeType = File::mimeType($path))) {
            throw new InvalidArgumentException("Could not determine mime type for {$path}");
        }
```
which returns `audio/x-wav`.

I'm not sure what the difference is, but I do see that OpenAI doesn't accept x-wav. I don't know if we should override the Media mimeType, or the OpenAI provider handling of this. 

As said, override the mimeType works, but I think more people will run into this

barryvdh:
Normalizes legacy formats to audio/wav

## Description
Normalizes the mimetype. For now, just for .wav files

Fixes https://github.com/prism-php/prism/issues/981

### DIFF
--- a/src/ValueObjects/Media/Media.php
+++ b/src/ValueObjects/Media/Media.php
@@ -253,7 +253,7 @@ class Media implements Arrayable
             $this->mimeType = (new finfo(FILEINFO_MIME_TYPE))->buffer($content) ?: null;
         }
 
-        $this->mimeType = match (strtolower($this->mimeType)) {
+        $this->mimeType = match ($this->mimeType) {
             'audio/x-wav', 'audio/wave', 'audio/x-pn-wav', 'audio/vnd.wave' => 'audio/wav',
             default => $this->mimeType,
         };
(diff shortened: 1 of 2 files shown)


barryvdh:
The gemini docs also state audio/wav:
https://ai.google.dev/gemini-api/docs/audio#supported-formats

PR #1022, opened 2026-05-30
docs: show OpenAI-compatible gateway setup

cerebrixos:
## Summary
- add an OpenAI-compatible gateway example to the OpenAI provider page
- show how to route Prism through Tuning Engines by changing the provider URL and API key
- keep Prism application calls unchanged with ->using('openai', 'gpt-4o')

## Why
We are a Rails and AI infrastructure team using Prism with Tuning Engines. Since Prism already supports configurable provider URLs, this gives Laravel/PHP teams a simple setup path for governed model access, policy checks, traces, audit logs, and usage/cost reporting without changing their Prism application logic. It would mean a lot to us if you are open to including it, and I am happy to revise the copy to fit the docs style.

## Validation
- git diff --check

### DIFF
--- a/docs/providers/openai.md
+++ b/docs/providers/openai.md
@@ -9,6 +9,33 @@
 ]
 ```
 
+### OpenAI-compatible gateways
+
+The OpenAI provider can also point at an OpenAI-compatible gateway by changing
+the configured URL and API key. This is useful for teams that want to keep their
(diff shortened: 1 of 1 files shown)


cerebrixos:
Hello team @pocketarc . Please do consider this PR for merge. So much appreciate in advance :)

PR #1029, opened 2026-07-12
feat(openai): support GPT-5.6 structured output

bradleybernard:
## Description

Adds native structured-output recognition for the official GPT-5.6 API model family:

- `gpt-5.6` (Sol alias)
- `gpt-5.6-sol`
- `gpt-5.6-terra`
- `gpt-5.6-luna`

Without these entries, `StructuredModeResolver` silently falls back to JSON mode, so Prism sends `json_object` instead of `json_schema` and ignores `schema.strict`.

OpenAI documents Structured Outputs support for each GPT-5.6 model:

- https://developers.openai.com/api/docs/models/gpt-5.6-sol
- https://developers.openai.com/api/docs/models/gpt-5.6-terra
- https://developers.openai.com/api/docs/models/gpt-5.6-luna

The regression test covers all four identifiers and asserts Prism sends `json_schema` with `strict: true`.

This is intentionally narrower than #991, which proposes broader prefix matching for versioned model names.

## Breaking Changes

None.

## Test plan

- `vendor/bin/pest tests/Providers/OpenAI/StructuredTest.php`
- `vendor/bin/pest` (1467 passed, 8 skipped)
- `vendor/bin/phpstan analyse --memory-limit=512M`
- Pint and Rector on changed files

### DIFF
--- a/src/Providers/OpenAI/Support/StructuredModeResolver.php
+++ b/src/Providers/OpenAI/Support/StructuredModeResolver.php
@@ -43,6 +43,10 @@ class StructuredModeResolver
             'gpt-5.1',
             'gpt-5.2',
             'gpt-5.4',
+            'gpt-5.6',
+            'gpt-5.6-sol',
+            'gpt-5.6-terra',
+            'gpt-5.6-luna',
         ]);
     }
 
--- a/tests/Providers/OpenAI/StructuredTest.php
+++ b/tests/Providers/OpenAI/StructuredTest.php
@@ -144,6 +144,48 @@ it('uses meta to define strict mode', function (): void {
     });
 });
 
+it('uses structured mode for GPT-5.6 models', function (string $model): void {
+    FixtureResponse::fakeResponseSequence(
+        'v1/responses',
+        'openai/strict-schema-setting-set'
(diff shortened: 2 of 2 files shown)


wishborn:
For context from a downstream fork — [Particle-Academy/prism](https://github.com/Particle-Academy/prism) already covers this, via **prefix matching** in [`StructuredModeResolver`](https://github.com/Particle-Academy/prism/blob/main/src/Providers/OpenAI/Support/StructuredModeResolver.php) rather than an exact-match list. It matches on the `gpt-5` family prefix, so `gpt-5.6` and the `-sol` / `-terra` / `-luna` variants already resolve to structured mode (sending `json_schema` with `strict`) without per-model entries — that's effectively the broader #991 approach this PR intentionally avoids.

Sharing in case it helps the narrow-vs-broad trade-off discussion: the prefix approach is zero-maintenance for new `gpt-5.x` snapshots, at the cost of assuming every future `gpt-5.*` supports structured output (which holds for the GPT-5.6 family per OpenAI's docs). The explicit list here is the safer, more deliberate side of that trade-off.

PR #1030, opened 2026-07-21
fix(audio): map webm and mp4 container mime types to audio extensions

mrmorgan-i:
## Description

`finfo` reports browser `MediaRecorder` output as `video/webm`, since WebM is a container format. That isn't in `GeneratesAudioFilename`'s match list, so it falls through to `default => 'mp3'` and the file gets uploaded as `audio.mp3` with WebM bytes inside it. OpenAI fails to parse it with:

```
OpenAI Error [400]: invalid_request_error - Audio file might be corrupted or unsupported
```

Same thing happens with `video/mp4`, which is what Safari's `MediaRecorder` produces.

This adds the container types to the existing match list. The `default => 'mp3'` fallback is unchanged, along with the three tests covering it.

Related to #981. That issue covers the `audio/x-wav` alias, and #1001 handles it at the `Media` layer, which makes sense for aliases that are for sure audio. `video/webm` is different because it's a truthful mime type, and normalizing it in `Media` would break genuine video input such as Gemini's. So this only touches the audio filename helper, and the two changes shouldn't conflict.

### DIFF
--- a/src/Concerns/GeneratesAudioFilename.php
+++ b/src/Concerns/GeneratesAudioFilename.php
@@ -11,13 +11,13 @@ trait GeneratesAudioFilename
         $extension = match ($mimeType) {
             'audio/flac' => 'flac',
             'audio/mpeg', 'audio/mp3' => 'mp3',
-            'audio/mp4' => 'mp4',
+            'audio/mp4', 'video/mp4' => 'mp4',
             'audio/mpga' => 'mpga',
             'audio/m4a', 'audio/x-m4a' => 'm4a',
-            'audio/ogg' => 'ogg',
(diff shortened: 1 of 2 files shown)


wishborn:
@mrmorgan-i — this is now released in the maintained fork ([`particle-academy/prism` v0.112.0](https://github.com/Particle-Academy/prism/releases/tag/v0.112.0)), which is a drop-in for `prism-php/prism`.

Your mapping went in as-is, plus one gap it didn't cover: `MediaRecorder` attaches codec parameters, so the very types this targets arrive as `audio/webm;codecs=opus` and still fell through to the mp3 default. The mime type is now reduced to its bare form before matching, so it works on real browser output.

Thanks for the fix.

PR #1032, opened 2026-07-25
Forward provider options in the DeepSeek handlers. Fixes #1031

### LINKED ISSUE #1031: Thinking mode on DeepSeek can't be disabled.
DeepSeek allows to disable thinking, but the parameter for it is not passed to the model.  Please check https://api-docs.deepseek.com/guides/thinking_mode

leobeal:
## Description
DeepSeek thinking mode option and others were simply being discarded. This PR adds support for them. 
FIxes #1031

### DIFF
--- a/src/Providers/DeepSeek/Handlers/Stream.php
+++ b/src/Providers/DeepSeek/Handlers/Stream.php
@@ -436,6 +436,13 @@ class Stream
                 'top_p' => $request->topP(),
                 'tools' => ToolMap::map($request->tools()) ?: null,
                 'tool_choice' => ToolChoiceMap::map($request->toolChoice()),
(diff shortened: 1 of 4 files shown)


wishborn:
@leobeal — this is now released in the maintained fork ([`particle-academy/prism` v0.112.0](https://github.com/Particle-Academy/prism/releases/tag/v0.112.0)), which is a drop-in for `prism-php/prism`.

Absorbed with one change: rather than the five-key allowlist, all three DeepSeek handlers now merge `providerOptions()` generically, so `reasoning_effort`, `stop`, `logprobs` and anything DeepSeek adds later work too. Options merge first, so nothing can clobber `model`/`messages`/`stream`. Your two test cases are in, plus a clobber guard.

Fixes the `thinking` case from #1031. Thanks for the fix.

PR #1033, opened 2026-07-31
fix: convert backed enum tool arguments

### LINKED ISSUE #1016: Feature request: auto-convert string values to BackedEnum instances when tool handler parameters are typed as enums
## Summary

When a tool parameter is declared with `withEnumParameter()` and the handler method types that parameter as a `BackedEnum` subclass, Prism forwards the raw string value from the model straight into `call_user_func`. PHP's strict typing rejects the call, the `TypeError` is caught, and the model receives a validation error instead of a successful tool invocation. I'd like Prism to cast the string to the enum before invoking the handler, analogous to Laravel's route-model binding of enums.

## Reproduction

```php
use Prism\Prism\Tool;

enum Classification: string
{
    case A = 'A';
    case B = 'B';
    case C = 'C';
}

class ClassifyTool extends Tool
{
    public function __construct()
    {
        $this->as('classify')
            ->for('Classify something')
            ->withEnumParameter(
                name: 'target',
                description: 'Target classification',
                options: array_map(fn (Classification $c) => $c->value, Classification::cases()),
            )
            ->using($this);
    }

    public function __invoke(Classification $target): string
    {
        return $target->value;
    }
}
```

The model sends `{"target": "B"}`. `"B"` is a string; `$target` expects `Classification`. The call fails.

## Current behavior

The `TypeError` is caught by `Tool::handle()` (around the `call_user_func($this->fn, ...$args)` line) and turned into a `ToolError` sent back to the model:

```
Parameter validation error: Type mismatch. Expected: [target (EnumSchema, required)]. Received: {"target":"B"}. Please provide correct parameter types and names.
```

The schema-declared type (`EnumSchema` with string options) and the PHP handler signature (`Classification`) disagree, and the gap is silently pushed onto the caller. Current workaround is to type the handler parameter as `string` and call `Classification::from(\$value)` manually inside the handler.

## Proposed behavior

Before calling the handler, inspect each handler parameter via reflection. If the type is a `BackedEnum` subclass and the corresponding argument is a string or int, convert it with `\$enumClass::from(\$value)`. If the schema for that parameter was declared with `withEnumParameter`, the developer has already told Prism the options map to enum cases — the cast is unambiguous.

```php
\$reflection = new ReflectionFunction(Closure::fromCallable(\$this->fn));

foreach (\$reflection->getParameters() as \$index => \$param) {
    \$type = \$param->getType();
    if (! \$type instanceof ReflectionNamedType) {
        continue;
    }

    \$typeName = \$type->getName();
    if (! is_subclass_of(\$typeName, BackedEnum::class)) {
        continue;
    }

    \$key = array_key_exists(\$param->getName(), \$args) ? \$param->getName() : \$index;
    \$value = \$args[\$key] ?? null;

    if (is_string(\$value) || is_int(\$value)) {
        \$args[\$key] = \$typeName::from(\$value);
    }
}
```

## Version and related work

- Reproduced on v0.99.22; `main` still calls `call_user_func` with raw args and does no handler-side casting.
- Searched existing issues/PRs for `enum`, `BackedEnum`, `withEnumParameter`, `tool parameter type`. Nothing overlapping. Closest tangentially-related open issue is #984 (Groq returns wrong wire types for boolean/number params), but that's provider-side normalization, distinct from handler-side enum casting.
- Prior enum-related PRs (#803, #395) move Prism toward broader `BackedEnum` support, so this fits the direction of travel.

Happy to open a PR with tests if the maintainers are open to the change.

mikemikimike:
## Summary

- Convert string/integer values returned for tool arguments into backed enum instances expected by the callable.
- Support both positional and named arguments.
- Surface invalid enum values as parameter validation errors.
- Add regression coverage for positional, named, and invalid values.

Closes #1016

## Validation

- `git diff --check` passed.
- Local Pest/PHPStan could not be run because this environment has no PHP/Composer installation and no `vendor` directory.
- GitHub Actions should run the repository's full Pest matrix and PHPStan checks.

### DIFF
--- a/src/Tool.php
+++ b/src/Tool.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Prism\Prism;
 
 use ArgumentCountError;
+use BackedEnum;
 use Closure;
 use Error;
 use Illuminate\Container\Container;
@@ -22,6 +23,8 @@ use Prism\Prism\Schema\StringSchema;
 use Prism\Prism\Tools\LaravelMcpTool;
 use Prism\Prism\ValueObjects\ToolError;
 use Prism\Prism\ValueObjects\ToolOutput;
+use ReflectionFunction;
+use ReflectionNamedType;
 use Throwable;
 use TypeError;
 
@@ -262,6 +265,7 @@ class Tool
     public function handle(...$args): string|ToolOutput|ToolError
     {
         try {
+            $args = $this->normalizeBackedEnumArguments($args);
             $value = call_user_func($this->fn, ...$args);
 
             if (is_string($value)) {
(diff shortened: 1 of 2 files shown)


mikemikimike:
Validation update:

- `git diff --check`: passed.
- Local Pest/PHPStan: not runnable in the current environment because PHP, Composer, and `vendor/` are unavailable.
- Full repository CI: the Formatting, PHPStan, and Tests workflow runs were created, but all are currently `action_required` because this fork PR requires maintainer approval before workflows execute. No full-context test result is being claimed yet.

The implementation covers positional and named backed-enum arguments, and converts invalid enum values into the existing parameter-validation error path.

mikemikimike:
Final local validation completed after installing PHP 8.3.33 and Composer 2.10.2 in a temporary project tool directory:

- Full Pest suite (single-process, `memory_limit=-1`): **1465 passed, 8 skipped, 4832 assertions**.
- PHPStan: **passed, no errors**.
- Pint on changed files (`src/Tool.php`, `tests/ToolTest.php`): **passed**.
- `git diff --check`: **passed**.

The initial parallel Pest run hit the Windows worker's default 128MB PHP memory limit in the existing ArchTest worker; rerunning the same full suite single-process completed successfully. The PR branch remains clean at commit `70a26a7f`.

